PR #5, opened 2018-08-21
Updating  Artifactory version to 6.2.0

jainishshah17:


stable/artifactory/templates/artifactory-statefulset.yaml
```diff
@@ -70,19 +75,16 @@ spec:
                 cp -Lrf /bootstrap/* /artifactory_extra_conf/;
                 {{- end }}
                 if [ -d {{ .Values.artifactory.persistence.mountPath }}/access/etc/keys ]; then
```
rimusz: remove:
                if [ -d {{ .Values.artifactory.persistence.mountPath }}/access/etc/keys ]; then
                fi;
which makes container crash

jainishshah17: @rimusz Fixed

Issue #19, opened 2018-08-29
wrong S3 binarystore configuration in artifactory-ha


bsemperlotti:
**Version of Helm and Kubernetes**:
NA

**Which chart**:
artifactory-ha

**What happened**:
The file `artifactory-binarystore.yaml` has a wrong hardcoded S3 provider configuration

```
2018-08-22 15:17:35,527 [eventual-cluster-worker-0] [ERROR] (o.j.s.b.p.RetryBinaryProvider:126) - Failed to check if blob 'ee0b58a1b6b0f5993e1ea6e155432e0152ba689a' exist in next binary provider
java.lang.RuntimeException: Failed to check if blob  '{sha1='ee0b58a1b6b0f5993e1ea6e155432e0152ba689a', sha2='null', md5='null', length=-1, headers={}}'  exist in s3: null - null
```

**What you expected to happen**:
When setting S3 backend for filestore, Artifactory should be able to connect to S3 to upload and retreive files.
With the current configutation, artifactory is unable to connect to S3

**How to reproduce it** (as minimally and precisely as possible):
Simply configure S3 has described by documentation

**Anything else we need to know**:

bsemperlotti:
S3 binarystore should be configured as described here:
https://jfrog.com/knowledge-base/how-to-setup-artifactory-ha-cluster-in-aws/

PR #20, opened 2018-08-29
Fix binarystore for S3 provider

bsemperlotti:
Fix endpoint and scheme values for AWS S3 provider

#19


bsemperlotti:
IMHO, bumping version this way is not a scalable solution as it greatly depends on the pull requests merge order.

rimusz:
any change to the chart needs version bump

PR #23, opened 2018-09-02
New Distribution. HA and non-root

eldada:
Add support for the new Distribution 1.2.0.
Two major changes:
1. Support HA (Distribution service)
2. Non-root Docker containers


jainishshah17:
@eldada @omerkay We will need to remove our current logic for creating configMap for the token. When we use this logic it adds newline char at end of file. Which changes checksum and gives an error in authenticating requests made by Distributor to Distribution.

jainishshah17:
I propose asking user to use `kubectl cp` command to move the token from Distribution to Distributor.

eldada:
@jainishshah17 - the issue with the newline in the token was fixed in the distribution code. @omerkay and I confirmed the new chart works as expected with token this way.

jainishshah17:
screen shot 2018-09-04 at 3 01 02 pm

jainishshah17:
I am still able to reproduce the token issue. Refer to above screenshot to see new line char at end of file.

stable/distribution/values.yaml
```diff
@@ -74,9 +74,9 @@ mongodb:
   mongodbExtraFlags:
   - "--wiredTigerCacheSizeGB=1"
   mongodbRootPassword:
+  mongodbDatabase: bintray
```
liorhson: mongo database name should be 'distribution'

eldada: @liorhson see https://github.com/jfrog/charts/pull/27

PR #25, opened 2018-09-02
Set Xray version 2.3.2

eldada:
Set Xray version 2.3.2


jainishshah17:
@eldada As we discussed please bumpup rabbitmq-ha version in `requirements.yaml`

jainishshah17:
@eldada Now we don't need `ci/test-values.yaml`. Can you please remove it.

stable/xray/values.yaml
```diff
@@ -111,12 +111,22 @@ mongodb:
 ##
 rabbitmq-ha:
   replicaCount: 1
-  rabbitmqUsername: "guest"
```
ankushchadha: I think the quotes were added for a reason, maybe this is not applicable to the new chart.

eldada: Yes. Was already tested with `guest` and works well.

PR #29, opened 2018-09-12
Distributor HA - one pod per services (DO NOT MERGE)

eldada:
**STILL WORK IN PROGRESS! Not ready for merge.**

A proposed solution for the true distribution HA architecture.
A single pod with all 3 containers:
- Distribution
- Distributor
- Redis (new version - `4.0.11-debian-9`)

Load balancing done on Distribution http endpoint only.
This is an identical architecture to the non-k8s HA design (one host with the 3 services).


eldada:
Not implementing.

PR #45, opened 2018-09-25
Adds a PodDisruptionBudget for the member nodes

valdisrigdon:
This ensures that during a rolling update of a cluster, one of the
member nodes will always be up serving traffic.


valdisrigdon:
`CHANGELOG.md` updated.

PR #56, opened 2018-09-27
chart: Set Artifactory/HA to 6.4.0

rimusz:
Updates Artifactory/HA charts to Artifactory v6.4.0


stable/artifactory-ha/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: artifactory-ha
 home: https://www.jfrog.com/artifactory/
-version: 0.4.7
-appVersion: 6.3.2
+version: 5.0.0
```
eldada: I think you meant `0.5.0`?

Issue #47, opened 2018-09-25
Adding nodeSelector to distribution and xray charts


sb464f:
**Is this a request for help?**: yes

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Feature Request

**Version of Helm and Kubernetes**: Helm 2.10.0 K8s 1.10

**Which chart**: Distribution and xray

**What happened**: Want node selector to specify the nodes in my my env

**What you expected to happen**: I want the pod to come up on specific node

**How to reproduce it** (as minimally and precisely as possible):

**Anything else we need to know**:

eldada:
Done with https://github.com/jfrog/charts/pull/60 and https://github.com/jfrog/charts/pull/61. Closing.

PR #60, opened 2018-09-30
Add nodeselector, affinity and tolerations for Distribution

eldada:
Add nodeSelector, affinity and tolerations for Distribution chart


eldada:
This is to fulfil https://github.com/jfrog/charts/issues/47

PR #72, opened 2018-10-09
Missing quote for ingress host names in all charts

### LINKED ISSUE #71: Missing quote for ingress host names in all charts

BUG REPORT

**Which chart**:

All charts

**What happened**:

Charts are not compatible with ingres hosts names begining with '*', Helm complains about syntax

Ingress host names should be quoted like this:

```
  {{- range $host := .Values.ingress.hosts }}
  - host: {{ $host | quote }}
```

See how the ingress template is created with "helm create" command.

sebastien-prudhomme:

Enable using host names beginings with '*' in ingress

**Which issue this PR fixes**

fixes #71 

As it is a minor change, i didn't update the changelog and the version of all charts

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All notable changes to this chart will be documented in this file.
 
+## [0.5.3] - Oct 9, 2018
+* Quote ingress hosts to support wildcard names
+
 ## [0.5.2] - Oct 2, 2018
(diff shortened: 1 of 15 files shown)


jainishshah17:
@sebastien-prudhomme Thank you for your contribution. Can you please update changelog as well.

rimusz:
also all changed charts version needs a bump too

eldada:
@sebastien-prudhomme 
It's ok, but needs CHANGELOG and version update for all charts.

sebastien-prudhomme:
ok, version and changelog updated

PR #74, opened 2018-10-11
Distribution version 1.3.0

eldada:
Distribution version 1.3.0


stable/distribution/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # JFrog Distribution Chart Changelog
```
rimusz: can you replace with: All changes to this chart will be documented in this file.

PR #75, opened 2018-10-11
Xray version 2.4.0

eldada:
Xray version 2.4.0


jainishshah17:
@eldada Can we change chart version to 0.6.0

PR #91, opened 2018-10-18
ingress annotation changes specifially to support docker registry

zendril:
@eldada @jainishshah17 
The current annotation support in the ingress templates iterates over items, which breaks certain concepts like configuration-snippets.

This enables the annotation section to be passed in verbatim (or omitted altogether).

No existing issue number, however it is tied to an internal jfrog support ticket.

By doing this, it introduces a warning from lint:
`2018/10/17 14:27:13 warning: destination for annotations is a table. Ignoring non-table value <nil>`

In my research (including talking to people on the #helm-users slack channel), this appears to be a known issue and a bug in Helm 2 which they tried to fix but rolled it back because it broke other dependencies like prometheus.

If you can find a way around this I would be happy to make the change and update the PR, but I'm currently not seeing one.

### DIFF
--- a/stable/artifactory-ha/README.md
+++ b/stable/artifactory-ha/README.md
@@ -394,6 +394,7 @@ The following table lists the configurable parameters of the artifactory chart a
 | `ingress.hosts`             | Artifactory Ingress hostnames       | `[]`                                                 |
 | `ingress.tls`               | Artifactory Ingress TLS configuration (YAML) | `[]`                                        |
 | `ingress.defaultBackend.enabled` | If true, the default `backend` will be added using serviceName and servicePort | `true` |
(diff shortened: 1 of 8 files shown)


zendril:
Done. 
You would think I'd remember to do it every time by now..  guess not :/

stable/artifactory-ha/README.md
```diff
@@ -495,6 +496,31 @@ Include the secret's name, along with the desired hostnames, in the Artifactory
           - artifactory.domain.com
 ```
 
+### Ingress annotations
```
jainishshah17: @zendril We should specify that these annotations are only to use Artifactory as Docker Registry with   Repository Path as a reverse proxy method.

jainishshah17:
@zendril Also please fix conflicts.

stable/artifactory-ha/README.md
zendril: Fair point. I'll update.

zendril: @jainishshah17 Done and pushed.

zendril:
> @zendril Also please fix conflicts.

@jainishshah17 Done and pushed.

PR #93, opened 2018-10-19
Enhancements to stable/artifactory-ha and stable/artifactory

robbie-demuth:
**What this PR does / why we need it**:

Appian is currently deploying a customized variant of the `stable/artifactory-ha` chart for production use. This PR incorporates changes we've needed to make to suit our CD workflow. See the changelog and commit history for more details. Some commits only apply to the `stable/artifactory-ha` chart. These commits are indicated by sentences in the commit message

PR #96, opened 2018-10-23
Fix external database secret

robbie-demuth:

This fixes a minor bug in the logic that was added to enable providing an external secret for external database credentials that was introduced by https://github.com/jfrog/charts/pull/93


robbie-demuth:
Could you all please review, @jainishshah17, @eldada, and @rimusz? Sorry I didn't catch this the first time around

PR #102, opened 2018-11-01
Adds support for arbitrary User Plugins

valdisrigdon:
See https://www.jfrog.com/confluence/display/RTF/User+Plugins for
examples on how to write User Plugins.

This adds support for providing arbitrary User Plugins deployed to Artifactory HA.

### DIFF
--- a/stable/artifactory-ha/templates/artifactory-plugins.yaml
+++ b/stable/artifactory-ha/templates/artifactory-plugins.yaml
@@ -1,13 +1,16 @@
 kind: ConfigMap
 apiVersion: v1
 metadata:
-  name: {{ template "artifactory-ha.fullname" . }}-isc
+  name: {{ template "artifactory-ha.fullname" . }}-plugins
   labels:
     app: {{ template "artifactory-ha.name" . }}
     chart: {{ template "artifactory-ha.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 data:
+{{- if .Values.artifactory.plugins }}
+{{ .Values.artifactory.plugins | toYaml | indent 2 -}}
+{{- end }}
   inactiveServerCleaner.groovy: |-
     import org.artifactory.state.ArtifactoryServerState
     import org.artifactory.storage.db.servers.service.ArtifactoryServersCommonService
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.7.2] - Nov 1, 2018
+* Allow specification of arbitrary [User Plugins](https://www.jfrog.com/confluence/display/RTF/User+Plugins)
+
 ## [0.7.1] - Oct 29, 2018
 * Change probes port to 8040 (so they will not be blocked when all tomcat threads on 8081 are exhausted)
 
--- a/stable/artifactory-ha/Chart.yaml
+++ b/stable/artifactory-ha/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: artifactory-ha
 home: https://www.jfrog.com/artifactory/
-version: 0.7.1
+version: 0.7.2
 appVersion: 6.5.2
 description: Universal Repository Manager supporting all major packaging formats,
   build tools and CI servers.
--- a/stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml
+++ b/stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml
@@ -27,6 +27,7 @@ spec:
         release: {{ .Release.Name }}
       annotations:
         checksum/binarystore: {{ include (print $.Template.BasePath "/artifactory-binarystore.yaml") . | sha256sum }}
+        checksum/plugins: {{ include (print $.Template.BasePath "/artifactory-plugins.yaml") . | sha256sum }}
       {{- range $key, $value := .Values.artifactory.annotations }}
         {{ $key }}: {{ $value | quote }}
       {{- end }}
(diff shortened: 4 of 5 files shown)


valdisrigdon:
Could you all please review, @jainishshah17, @eldada, and @rimusz?

stable/artifactory/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: artifactory
 home: https://www.jfrog.com/artifactory/
-version: 7.7.0
+version: 0.7.1
```
jainishshah17: @valdisrigdon Looks like you updated the plugin to `artifactory-ha` and bumped version of `artifactory` chart. We like to keep both our charts in sync. Can you add this changes in both charts?

valdisrigdon: I looked at doing this, but since User Plugins looks like an Artifactory Pro feature, I didn't add it to the non-HA version.

stable/artifactory-ha/templates/artifactory-plugins.yaml
```diff
@@ -1,13 +1,16 @@
...
   labels:
     app: {{ template "artifactory-ha.name" . }}
     chart: {{ template "artifactory-ha.chart" . }}
     heritage: {{ .Release.Service }}
     release: {{ .Release.Name }}
 data:
+{{- if .Values.artifactory.plugins }}
+{{ .Values.artifactory.plugins | toYaml | indent 2 -}}
+{{- end }}
   inactiveServerCleaner.groovy: |-
```
valdisrigdon: This doesn't exist in the `artifactory` chart.

jainishshah17:
@valdisrigdon How about just taking the configMap name from `values.yaml` (instead of the entire plugin code) and mount it to Artifctory pods. This way user will be able to add one configMap with all files related to the plugin or multiple plugins and mount it to Artifactory pods.

WDYT @rimusz @eldada?

eldada:
I agree with @jainishshah17 . It will help users manage all their plugins as proper groovy files and just create a ConfigMap from them. We should also do it for the `inactiveServerCleaner.groovy` plugin.

valdisrigdon:
I think I would prefer that solution as well, but there are two downsides right now:

* the `entrypoint` of the container will copy from `/tmp/plugins` to load them, but if there are two `ConfigMap`s, you can't mount both to `/tmp/plugins`. This would require a change to the docker container entrypoint as well.
* by including them in the existing `ConfigMap` I can generate the checksum which would force a rolling update to push out the new plugins.

Thoughts on this? This first point above is a blocker right now with the approach you are suggesting I believe.

eldada:
For the first issue, isn't it possible to mount multiple `ConfigMap`s to the same directory using the `subPath` option?
I like the idea of the checksum., We can figure out other ideas.

valdisrigdon:
@eldada You can use `subPath` but that requires you to know the names of the entries in the `configMap`. Would you want the caller to pass in the name of the `ConfigMap` plus each groovy filename? I would expecting to pass just the `configMap` name and have all the entries applied as plugins.

eldada:
@valdisrigdon - we are still discussing this internally. Sorry for the wait.

zendril:
@valdisrigdon @eldada Halfway reading through this but we are also need to load user plugins into artifactory-ha from helm. Is the sticking point how to get multiple config maps to all land appropriately in HOME/etc/plugins?

At a minimum could load each as a configmap and then use the preStart hoook that is now in the *-statefulset.yamls to copy them to either /tmp/plugins or right into HOME/etc/plugins, similar to what is happening for the bootstrap config?

valdisrigdon:
We build a custom docker image for Artifactory, so we just load them at build time right now. We are trying to move away from that, hence this feature request.

eldada:
I'd like the plugins scripts sources **not** to be managed as the part of the chart's sources, but as an actual separate source code repository.
The entrypoint script has the following:
```
addPlugins () {
    logger "Adding plugins if exist"
    cp -fv /tmp/plugins/* ${ARTIFACTORY_HOME}/etc/plugins/
}
```
So we just need to pre-load the plugins to `/tmp/plugins`.
I think we should change the PR to the following behaviour:
1. User creates one configMap per plugin from the plugins sources
2. `Values.yaml` include a key with array of configMaps
```
artifactory:
  pluginsConfigMaps:
  - name: pluginConfigMap1
  - name: pluginConfigMap2
```
3. The `artifactory-statefulset.yaml` has a conditional loop to mount each plugin to `/tmp/plugins` (`volumes` and `volumeMounts`)

WDYT? @jainishshah17 ? @rimusz ? @valdisrigdon ?

valdisrigdon:
Could the `ConfigMap` support multiple plugins? It seems like it could support many `ConfigMap`s with many plugins in each.

+1 to the user creating the `ConfigMap`. I would much prefer creating a single `ConfigMap` that contains all of my plugins and then passing that as a reference to the artifactory chart.

eldada:
Yes. A single `ConfigMap` can be created from multiple plugins (files). But, need to see and test how it mounts to `/tmp/plugins` and that the `cp` in the entrypoint picks it up properly. 
Can you try and share the changes in this PR?

zendril:
@valdisrigdon Let me know if you either want help and or testing it when you get something. We are actively working on the same on our side as of yesterday.

valdisrigdon:
@zendril I won't have time to get to this in the near future, especially as we have a workaround for the moment. I'd be happy to test out what you have though if you push something up!

zendril:
@valdisrigdon Ok. We have a working version of it locally now using the config maps, etc.. . I'll do a review on my side and probably create a new PR in a day or so.

zendril:
FYI, Still doing a few more tweaks on our side to remove any hard-coding specific to our usage.

jainishshah17:
https://github.com/jfrog/charts/pull/278/

jainishshah17:
I am using secret instead of configMap for security purpose as some of our User plugins have credentials/tokens in configuration files.

jainishshah17:
I also tried using multiple secrets per plugin. In that case, mounting all plugin in `/tmp/plugins` becomes challenging.

zendril:
@jainishshah17 Is this defunct now that https://github.com/jfrog/charts/pull/278 is merged?

PR #106, opened 2018-11-07
Support Artifactory db url

eldada:
When not using the PostgreSQL sub-chart, add support for setting `DB_URL` with `database.url`.


jainishshah17:
@eldada Any reason for not making `.Values.database.type` required for `artifactory` chart?

eldada:
@jainishshah17 - an empty `DB_TYPE` defaults to running with derby. So no need to force it in `artifactory`. Only in `artifactory-ha`

PR #110, opened 2018-11-12
Document existingClaim in Artifactory Chart

m0xx:
add `artifactory.persistence.existingClaim` to values table in README.


m0xx:
Sorry, doesn't work as expected, existingClaim is a boolean to activate claim selector.

PR #112, opened 2018-11-12
Allow pod anti-affinity settings to include primary node

### LINKED ISSUE #111: Allow pod anti-affinity settings to include primary node

FEATURE REQUEST

Currently, pod anti-affinity applies only to member nodes, which may result in a member pod being scheduled onto the same host where primary node is already running. In a case when we have only two nodes, primary and member, such situation will cause a complete unavailability of Artifactory service. A solution could be to avoid limiting poddisruptionbugdet to node role "member", like it is alredy done with service definition here: 
https://github.com/jfrog/charts/blob/master/stable/artifactory-ha/templates/artifactory-service.yaml#L20-L22

ivanfitenko:
Allow pod anti-affinity settings to include primary node 
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #111

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,7 +1,10 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
-## [0.7.1] - Oct 29, 2018
+## [0.7.2] - Oct 29, 2018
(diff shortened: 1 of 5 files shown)


eldada:
@ivanfitenko - please update `Chart.yaml` with new version and also add changes to `CHANGELOG.md`

PR #114, opened 2018-11-12
Optionally include primary node into poddisruptionbudget

### LINKED ISSUE #113: Optionally include primary node into poddisruptionbudget

FEATURE REQUEST

When primary node is added to member service (with "pool: all" instead of "pool: members" in values.yaml) it should also be added to poddisruptionbudget to prevent a situation when there is only two nodes (primary and member-0) and both are deleted at the same time during eviction or similar operation

ivanfitenko:
Optionally include primary node into poddisruptionbudget

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #113


jainishshah17:
@ivanfitenko Thank you for your contribution. Can you please bump up chart version and also update Changelog and Readme accordingly?

eldada:
Also, please elaborate on the solution. What problem are you trying to solve?

ivanfitenko:
@eldada 
I have two artifactory nodes, one of them is primary, and the second is a member node. In values.yaml, I have "artifactory.service.pool" set to "all" so that both nodes can serve requests. In order to prevent some voluntary disruptions (e.g. node eviction) from bringing down one of my nodes while the other one is down, I want to include both primary and member node into PodDisruptionBugdet.

ivanfitenko:
@jainishshah17
No changes to readme are required. This pull request does not add any explicitly configurable options, it just make PodDisruptionBudget's behavior consistent with Service settings when "artifactory.service.pool" value is set to "all".

eldada:
@ivanfitenko - can you rebase and test this so we can merge it?

PR #150, opened 2018-12-16
Distribution version 1.4.0

eldada:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Upgrade Distribution to 1.4.0
- Migration to PostgreSQL as database
- Upgrade by deploying MongoDB and PostgreSQL (`--set mongodb.enabled=true`)

### DIFF
--- a/stable/distribution/templates/database-secrets.yaml
+++ b/stable/distribution/templates/database-secrets.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.database.password }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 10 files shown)


stable/distribution/templates/database-secrets.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
jainishshah17: @eldada Can you add condition check here. Create secret only if values are provided in .`Values.database` and postgress and mondo are disabled.

jainishshah17:
@eldada Does this support upgrade from 1.3.0 to 1.4.0? If no then bump up major version of chart.

jainishshah17:
@eldada Thank you for making those changes.

PR #174, opened 2019-01-08
Make artifactory.replicator.publicUrl required when the replicator is…

danielezer:
… enabled

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

<!--

Currently, when artifactory.replicator.enabled is set to true, the artifactory.replicator.publicUrl is required for proper configuration of the replicator but when running helm install there is no indication of that and its hard to know that the parameter is missing. This will fail the chart installation/upgrade when the replicator is enabled but the public url is not set.


danielezer:
@eldada please review :)

danielezer:
@jainishshah17 Done

PR #184, opened 2019-01-22
Add support for Artifactory extra custom init containers

eldada:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Adding support for extra custom init containers that are useful for special dedicated bootstrapping.

**Note**: Once this PR is approved and merged, we'll add this to the other charts as well.

### DIFF
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -125,7 +125,7 @@ artifactory:
   #      - 'touch {{ .Values.artifactory.persistence.mountPath }}/example-custom-setup'
   #    volumeMounts:
   #      - mountPath: "{{ .Values.artifactory.persistence.mountPath }}"
-  #        name: artifactory-volume
+  #        name: volume
 
   ## Artifactory requires a unique master key
   ## You can generate one with the command:
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -4,7 +4,7 @@
 # Access the values with {{ .Values.key.subkey }}
 
 # Common
-initContainerImage: "alpine:3.6"
+initContainerImage: "alpine:3.8"
 
 # For supporting pulling from private registries
 imagePullSecrets:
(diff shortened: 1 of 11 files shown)


stable/artifactory-ha/values.yaml
```diff
@@ -113,6 +113,20 @@ artifactory:
...
+  #  - name: "custom-setup"
+  #    image: "{{ .Values.initContainerImage }}"
+  #    imagePullPolicy: "{{ .Values.artifactory.image.pullPolicy }}"
+  #    command:
+  #      - 'sh'
+  #      - '-c'
+  #      - 'touch {{ .Values.artifactory.persistence.mountPath }}/example-custom-setup'
+  #    volumeMounts:
+  #      - mountPath: "{{ .Values.artifactory.persistence.mountPath }}"
+  #        name: artifactory-volume
```
jainishshah17: @eldada I think value of this should be just `volume` https://github.com/jfrog/charts/blob/09878bed4b355e93b35433ecb0a712db2168c2ee/stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml#L214

stable/artifactory/values.yaml
```diff
@@ -66,6 +66,19 @@ artifactory:
...
+  #  - name: "custom-setup"
+  #    image: "{{ .Values.initContainerImage }}"
+  #    imagePullPolicy: "{{ .Values.artifactory.image.pullPolicy }}"
+  #    command:
+  #      - 'sh'
+  #      - '-c'
+  #      - 'touch {{ .Values.artifactory.persistence.mountPath }}/example-custom-setup'
+  #    volumeMounts:
+  #      - mountPath: "{{ .Values.artifactory.persistence.mountPath }}"
+  #        name: artifactory-volume
```
jainishshah17: Same here
https://github.com/jfrog/charts/blob/09878bed4b355e93b35433ecb0a712db2168c2ee/stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml#L214

stable/artifactory-ha/values.yaml
eldada: Fixed

stable/artifactory/values.yaml
eldada: Here it's right. No need to fix.

PR #201, opened 2019-02-05
Adds support to customize terminationGracePeriodSeconds for Artifactory

valdisrigdon:
#### PR Checklist
- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated

This allows us to change the `terminationGracePeriodSeconds` which is now always the default of 30s. We occasionally send large files through Artifactory, such as custom Gradle distributions, and these take a while to download.

The pre-built docker container `docker.bintray.io/jfrog/artifactory-pro:6.7.0` doesn't allow modification of the timeout to Tomcat's stop which is 5s. However, we are using a custom docker container that customizes `artifactory.sh` to call `stop` to increase that timeout.  I don't know where that repository is, but I can make a Pull Request to get that change in as well.

The other option would be to add a `preStop` that would trigger the stop call to Tomcat; this would allow for the Tomcat timeout to be specified via the Helm Chart, and rely on the existing 5s timeout as a fail-safe. Thoughts on this?


valdisrigdon:
@jainishshah17 @eldada @rimusz Can I get a review on this?

PR #206, opened 2019-02-07
feat(artifactory-ha): Add logging sidecars

k:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Currently artifactory does output all logs to standard out, and some users may want more logs to be accessible from the kubernetes API. This PR allows the addition of tailing sidecars to grab log data and output them to stdout for consumption by the kubernetes API. The benefit of doing it in separate side cars is that for log collectors like fluentd who collect metadata about the pod it becomes easier to filter by which container the log is coming from.


stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -75,6 +75,18 @@ spec:
...
     {{- end }}
       containers:
+      {{- range .Values.artifactory.loggers }}
+      - name: {{ . | replace "_" "-" }}
+        image: busybox
+        command:
+        - tail
+        args:
+        - "-f"
+        - "/data/logs/{{ . }}.log"
```
jainishshah17: @k Do you mean `/var/opt/jfrog/artifactory/logs/` here?

jainishshah17: If that's the case replace it with `{{ .Values.artifactory.persistence.mountPath }}/logs/{{ .}}.log`

stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml
```diff
@@ -75,6 +75,18 @@ spec:
...
     {{- end }}
       containers:
+      {{- range .Values.artifactory.loggers }}
+      - name: {{ . | replace "_" "-" }}
+        image: busybox
+        command:
+        - tail
+        args:
+        - "-f"
+        - "/data/logs/{{ . }}.log"
```
jainishshah17: @k Same Here

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -75,6 +75,18 @@ spec:
 {{ tpl .Values.artifactory.customInitContainers . | indent 6}}
     {{- end }}
       containers:
+      {{- range .Values.artifactory.loggers }}
```
eldada: Can you move this to be after the main Artifactory container?

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -75,6 +75,18 @@ spec:
...
+      - name: {{ . | replace "_" "-" }}
+        image: busybox
```
eldada: Please get the image from a parameter. Also set a value with a specific tag. We don't like `latest` as a default.

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -75,6 +75,18 @@ spec:
...
+        command:
+        - tail
+        args:
+        - "-f"
```
eldada: Better use `-F` so it will wait for logs that were not created yet.

stable/artifactory-ha/values.yaml
```diff
@@ -114,6 +114,15 @@ artifactory:
...
 
+  loggers: []  # add any of the loggers to a sidecar if you want to be able to see them with kubectl logs or a log collector in your k8s cluster
+  # - request
+  # - event
+  # - binarystore
+  # - request_trace
+  # - access
+  # - artifactory
+  # - build_info_migration
+
```
eldada: Can you also add the tomcat logs? They are under the `logs/catalina` directory, and this feature will be very useful for collecting them too.
For this you will also need to update the `{{ . | replace "_" "-" }}` to care for `/`.

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
k: Ah I was wondering why my logs stopped when I checked today. Thanks for this 👍

stable/artifactory-ha/values.yaml
k: The problem with this is there does not appear to be a fixed file name where the most recent logs are. When I ls in the container I see:

```
/data/logs/catalina # ls
catalina.2019-02-07.log      catalina.2019-02-09.log      host-manager.2019-02-08.log  localhost.2019-02-07.log     localhost.2019-02-09.log     manager.2019-02-08.log
catalina.2019-02-08.log      host-manager.2019-02-07.log  host-manager.2019-02-09.log  localhost.2019-02-08.log     manager.2019-02-07.log       manager.2019-02-09.log
```

k: We could do tail -f catalina.*.log, but that will make the line-level output messy when the file changes, also unsure whether it will pick up new files.

eldada: You are right. It will not pick up new files. Only existing. We'll try to improve it on our side...

PR #260, opened 2019-03-21
[artifactory] Make ingress path configurable

stevie-:
Depending on ingress used, path is interpreted as / or wildcard /*.

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

GCE ingress or AWS ALB ingress interpret ingress path differently as f.e. nginx ingress.

### DIFF
--- a/stable/artifactory-ha/templates/ingress.yaml
+++ b/stable/artifactory-ha/templates/ingress.yaml
@@ -29,7 +29,7 @@ spec:
   - host: {{ $host | quote }}
     http:
       paths:
-        - path: /
+        - path: {{ .Values.ingress.path }}
           backend:
             serviceName: {{ $serviceName }}
             servicePort: {{ $servicePort }}
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.11.14] - Mar 21, 2019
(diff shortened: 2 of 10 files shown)


stevie-:
No idea why helm install times out in Build test.

eldada:
@stevie- it might be related to Bintray issues, since our images are pulled from there. I initiated a new build.

stevie-:
@eldada thanks

stable/artifactory-ha/templates/ingress.yaml
```diff
@@ -29,7 +29,7 @@ spec:
   - host: {{ $host | quote }}
     http:
       paths:
-        - path: /
+        - path: {{ .Values.ingress.path }}
```
jaspermarcus: Should this have been `$.Values.ingress.path` instead? I'm trying to upgrade to `0.11.14` and have ran into an issue around this templating. As I understand it, the `.` no longer points to the root context inside a range function (https://github.com/helm/helm/issues/1311#issuecomment-399577623)

eldada: @stevie- can you verify this?

eldada: @jaspermarcus - we are fixing it with https://github.com/jfrog/charts/pull/265.

stevie-: Ahh, Thanks, missed that. Guess tests coverage needs to be udpated as well.

jaspermarcus: Thanks for the quick response! 👍

PR #265, opened 2019-03-22
Fix the issue in the ingress path caused by using . inside of a range…

danielezer:
… clause instead of $

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Fix the issue with the ingress path - https://github.com/jfrog/charts/pull/260/files#r267981799

fixes #260 

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.11.16] - Mar 21, 2019
+* Fix issue with ingress path
+
 ## [0.11.15] - Mar 21, 2019
 * Added support for `artifactory.customSidecarContainers` to create custom sidecar containers
 * Added support for `artifactory.customVolumes` to create custom volumes
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,8 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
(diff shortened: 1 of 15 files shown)


jaspermarcus:
Thanks for fixing this so quickly. Is this fix meant to be available as part of chart `0.11.16`? I just tried pulling down the latest chart this morning, and it seemed to contain the readinessProbe changes that also bumped the chart version to `0.11.16`, but it seemed to be missing this change.

rimusz:
two releases went with the same chart version `0.11.16`, the last one has this fix

rimusz:
looks like it jenkins did not finish the job, triggered it again, now the chart has changes from this PR, sorry for confusion

zendril:
Using helm template, this breaks ingress now with the following:

Error: render error in "artifactory-ha/templates/ingress.yaml": template: artifactory-ha/templates/ingress.yaml:32:26: executing "artifactory-ha/templates/ingress.yaml" at <.Values.ingress.path>: can't evaluate field Values in type interface {}

I can recreate it by setting `enabled: true` and uncommenting the hosts entry `- artifactory.domain.example`
```
ingress:
  enabled: true
  defaultBackend:
    enabled: true
  # Used to create an Ingress record.
  hosts:
    - artifactory.domain.example
```

Currently using helm 2.9.1, was this '. in a control structure scope change' changed in a newer version of helm perhaps?

I'll upgrade and do some testing, but might be worth calling out in the release notes or readme that this chart requires a minimum version of helm. Probably a decent thing to have in general anyways?

zendril:
Confirmed. 2.13.1 works. 

@rimusz I see we have 
```
## Prerequisites Details

* Kubernetes 1.8+
* Artifactory HA license
```
At the top of our README.md.. does it make sense to also call out minimum helm version as well?

PR #273, opened 2019-03-27
Mission control 3.5.0

amithins:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Update Mission Control to 3.5.0 
HA support
Move Elasticsearch inside Mission Control pod
Upgrade Elasticsearch to 6.6.0


jainishshah17:
@amithins can you please fix CI failure?

stable/mission-control/values.yaml
```diff
@@ -99,7 +58,7 @@ postgresql:
   enabled: true
   imageTag: "9.6.11"
   postgresUsername: postgres
-  postgresPassword:
+  postgresPassword: postgres
```
eldada: We should not hardcode a default password here. Let it be auto generated or passed as a parameter or a secret. 
Also need to document that in case of allowing auto generation, the following `helm upgrade` command must include this password so not to generate a new one.

stable/mission-control/values.yaml
```diff
@@ -175,26 +134,47 @@ database:
...
+    size: 100Gi
+    ## ElasticSearch data Persistent Volume Storage Class
+    ## If defined, storageClassName: <storageClass>
+    ## If set to "-", storageClassName: "", which disables dynamic provisioning
+    ## If undefined (the default) or set to null, no storageClassName spec is
+    ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
+    ##   GKE, AWS & OpenStack)
+    ##
+    # storageClass: "-"
+
```
eldada: Need a resources block
```
  resources: {}
  #  requests:
  #    memory: "2Gi"
  #    cpu: "250m"
  #  limits:
  #    memory: "2.5Gi"
  #    cpu: "1"
```

stable/mission-control/values.yaml
rimusz: +1

stable/mission-control/values.yaml
rimusz: +1

rimusz:
TravisCI has issues with VMs booting https://www.traviscistatus.com
Screenshot 2019-03-28 17 03 29

PR #278, opened 2019-03-29
Adding support for user plugins in Artifactory

jainishshah17:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/stable/artifactory-ha/README.md
+++ b/stable/artifactory-ha/README.md
@@ -425,7 +425,7 @@ artifactory:
 ### Add Artifactory User Plugin during installation
 If you need to use [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins), you can use this options.
 
-Create secret with [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins) by following command:
+Create a secret with [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins) by following command:
 ```
 # Secret with single user plugin
 kubectl  create secret generic archive-old-artifacts --from-file=archiveOldArtifacts.groovy --namespace=artifactory-ha 
@@ -442,7 +442,7 @@ artifactory:
     - webhook
 ```
 
-You can now pass above created `plugins.yaml` file helm install command to deploy Artifactory with user plugins as follows:
+You can now pass the created `plugins.yaml` file to helm install command to deploy Artifactory with user plugins as follows:
 ```
 helm install --name artifactory-ha -f plugins.yaml jfrog/artifactory-ha
 ```
@@ -468,7 +468,7 @@ The following table lists the configurable parameters of the artifactory chart a
 | `artifactory.customInitContainers`| Custom init containers                  |                                            |
 | `artifactory.customSidecarContainers`| Custom sidecar containers            |                                            |
 | `artifactory.customVolumes`       | Custom volumes                    |                                                  |
-| `artifactory.userPluginSecrets`   | Secret name array for Artifactory user plugins |                                     |
+| `artifactory.userPluginSecrets`   | Array of secret names for Artifactory user plugins |                                 |
 | `artifactory.masterKey`           | Artifactory Master Key. Can be generated with `openssl rand -hex 32` |`FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF`|
 | `artifactory.masterKeySecretName` | Artifactory Master Key secret name                     |                             |
 | `artifactory.accessAdmin.password`               | Artifactory access-admin password to be set upon startup|             |
--- a/stable/artifactory-ha/README.md
+++ b/stable/artifactory-ha/README.md
@@ -428,18 +428,23 @@ If you need to use [Artifactory User Plugin](https://github.com/jfrog/artifactor
 Create secret with [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins) by following command:
 ```
 # Secret with single user plugin
-kubectl  create secret generic plugins  --from-file=archiveOldArtifacts.groovy --namespace=artifactory-ha 
+kubectl  create secret generic archive-old-artifacts --from-file=archiveOldArtifacts.groovy --namespace=artifactory-ha 
 
 # Secret with single user plugin with configuration file
-kubectl  create secret generic plugins  --from-file=webhook.groovy  --from-file=webhook.config.json.sample --namespace=artifactory-ha
-
-# Secret with multiple user plugin
-kubectl  create secret generic plugins  --from-file=webhook.groovy  --from-file=webhook.config.json.sample --from-file=archiveOldArtifacts.groovy --from-file=buildCleanup.groovy --from-file=buildCleanup.properties --namespace=artifactory-ha
+kubectl  create secret generic webhook --from-file=webhook.groovy  --from-file=webhook.config.json.sample --namespace=artifactory-ha
 ```
 
-You can now pass above create secret with helm install command as follows:
+Add plugin secret names to `plugins.yaml` as following:
+```yaml
+artifactory:
+  userPluginSecrets:
+    - archive-old-artifacts
+    - webhook
 ```
-helm install --name artifactory-ha --set artifactory.userPluginsSecret=plugins jfrog/artifactory-ha
+
+You can now pass above created `plugins.yaml` file helm install command to deploy Artifactory with user plugins as follows:
+```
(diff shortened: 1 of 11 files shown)


stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -330,6 +334,11 @@ spec:
       - name: binarystore-xml
         configMap:
           name: {{ template "artifactory-ha.fullname" . }}-bs
+      {{- if .Values.artifactory.userPluginsSecret }}
+      - name : user-plugins
+        secret:
+          secretName: {{ .Values.artifactory.userPluginsSecret }}
```
robbie-demuth: Would you be amenable to passing this through `tpl`? For our use case, we actually use Artifactory as a sub chart dependency of a 'wrapper' chart. This means that the name of the secret is actually dynamic based on the name of the Helm release.

robbie-demuth: The same goes for all of the other StatefulSets as well obviously

jainishshah17: Will take a look

properpanda:
I dont see how the plugin gets to {ARTIFACTORY_HOME}/etc/plugins?  only that the secret gets mounted to /tmp/plugins.   Are we responsible for 'installing' the plugin? am i missing something?

zendril:
Adding @properpanda who is working on the same features on our side.

properpanda:
> I dont see how the plugin gets to {ARTIFACTORY_HOME}/etc/plugins? only that the secret gets mounted to /tmp/plugins. Are we responsible for 'installing' the plugin? am i missing something?

It appears that entrypoint-artifactory.sh does this copy automatically....

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
jainishshah17: @robbie-demuth We are not creating secret as part of helm deployment. I think you should be able to pass secret names using `--set` or through `values.yaml` file during helm installation.

stable/artifactory-ha/README.md
```diff
@@ -422,6 +422,31 @@ artifactory:
...
+
+Add plugin secret names to `plugins.yaml` as following:
+```yaml
+artifactory:
+  userPluginSecrets:
+    - archive-old-artifacts
+    - webhook
+```
+
+You can now pass above created `plugins.yaml` file helm install command to deploy Artifactory with user plugins as follows:
```
eldada: Should be "You can now pass the created `plugins.yaml` file to `helm install`"

stable/artifactory-ha/README.md
```diff
@@ -422,6 +422,31 @@ artifactory:
     ## Custom volume comes here ##
 ```
 
+### Add Artifactory User Plugin during installation
+If you need to use [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins), you can use this options.
+
+Create secret with [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins) by following command:
```
eldada: Should be "Create a secret"

stable/artifactory-ha/README.md
```diff
@@ -443,6 +468,7 @@ The following table lists the configurable parameters of the artifactory chart a
 | `artifactory.customInitContainers`| Custom init containers                  |                                            |
 | `artifactory.customSidecarContainers`| Custom sidecar containers            |                                            |
 | `artifactory.customVolumes`       | Custom volumes                    |                                                  |
+| `artifactory.userPluginSecrets`   | Secret name array for Artifactory user plugins |                                     |
```
eldada: Should be "Array of secret names for Artifactory user plugins"

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
robbie-demuth: For our use case, that is not possible. We deploy a chart that marks this chart as a dependency, so the secret _is_ created in the same Helm release as StatefulSets for Artifactory. The name of the secret is also based on the name of the Helm release since it is something like `name: {{ template "artifactory-ha-wrapper.fullname" . }}`. It would be great if this line could call `tpl` so that we can pass `{{ template "artifactory-ha-wrapper.fullname" . }}` as a value and have the inner Artifactory chart resolve the templating language

robbie-demuth: CC @valdisrigdon

stable/artifactory-ha/README.md
jainishshah17: @eldada done

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
eldada: @robbie-demuth - I suggest you submit a PR with support for your method once this is merged. We welcome features that will help our charts users.

stable/artifactory-ha/README.md
```diff
@@ -422,6 +422,26 @@ artifactory:
     ## Custom volume comes here ##
 ```
 
+### Add Artifactory User Plugin during installation
+If you need to use [Artifactory User Plugin](https://github.com/jfrog/artifactory-user-plugins), you can use this options.
```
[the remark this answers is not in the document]
rimusz: please address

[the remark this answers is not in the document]
jainishshah17: done

stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml
```diff
@@ -84,6 +84,11 @@ spec:
         - '/bin/sh'
         - '-c'
         - >
+          {{- if .Values.artifactory.userPluginSecrets }}
+          echo "Copying plugins";
+          mkdir -p {{ .Values.artifactory.persistence.mountPath }}/etc/plugins;
+          cp -Lrf /tmp/plugin/*/* {{ .Values.artifactory.persistence.mountPath }}/etc/plugins/;
```
robbie-demuth: @jainishshah17 - I think this actually breaks support for providing the cluster license via a Kubernetes secret. The `entrypoint-artifactory.sh` file contains the following logic that only copies files from the `/artifactory_extra_conf` directory into `$ARTIFACTORY_DATA/etc/` if the latter doesn't exist:

```bash
    if [ ! -d ${ARTIFACTORY_DATA}/etc ]; then
        mkdir -pv ${ARTIFACTORY_DATA}/etc || errorExit "Failed creating ${ARTIFACTORY_DATA}/etc"

        # Add extra conf files to a newly created etc/ only!
        addExtraConfFiles
    fi
```

Since this effectively puts the plugins in that directory, the license isn't copied

robbie-demuth: Here are the relevant logs from our primary node:

```
2019-04-03 15:19:07  [262 entrypoint-artifactory.sh] Setting up Access data directories if missing
2019-04-03 15:19:07  [152 entrypoint-artifactory.sh] Adding extra configuration files to /var/opt/jfrog/artifactory/access/etc if any exist
```

As you can see, it skips over the code block below that's intended to copy the files from `/artifactory_extra_conf`:

```
# Add additional conf files that were mounted to ARTIFACTORY_EXTRA_CONF
addExtraConfFiles () {
    logger "Adding extra configuration files to ${ARTIFACTORY_HOME}/etc if any exist"

    # If directory not empty
    if [ -d "${ARTIFACTORY_EXTRA_CONF}" ] && [ "$(ls -A ${ARTIFACTORY_EXTRA_CONF})" ]; then
        logger "Adding files from ${ARTIFACTORY_EXTRA_CONF} to ${ARTIFACTORY_HOME}/etc"
        cp -rfv ${ARTIFACTORY_EXTRA_CONF}/* ${ARTIFACTORY_HOME}/etc || errorExit "Copy files from ${ARTIFACTORY_EXTRA_CONF} to ${ARTIFACTORY_HOME}/etc failed"
    fi
}
```

robbie-demuth: CC @eldada @rimusz

zendril: @eldada @rimusz @robbie-demuth Check out https://github.com/jfrog/charts/pull/289

We noticed this breaks a bunch of stuff for a fresh install because of the exact check you pointed out.

@properpanda submitted https://github.com/jfrog/charts/pull/289 and issue https://github.com/jfrog/charts/issues/287 for it.

PR #285, opened 2019-04-02
Allow secret names for user plugins to contain template language

robbie-demuth:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Our organization deploys a Helm chart that actually has a dependency on the `artifactory-ha` chart. Now that the Artifactory chart supports user plugins, we'll create the secrets containing the user plugins in our 'wrapper' chart. Since the names of the secrets are named according to the output of template functions, we don't have a hardcoded secret name that we can pass to Artifactory. This PR updates the Artifactory chart to allow the names of externally provided secrets to contain template language

FYI @jainishshah17 @eldada - This is the change I was suggesting in https://github.com/jfrog/charts/pull/278#discussion_r270626240

### DIFF
--- a/stable/artifactory-ha/README.md
+++ b/stable/artifactory-ha/README.md
@@ -499,6 +499,14 @@ You can now pass the created `plugins.yaml` file to helm install command to depl
 helm install --name artifactory-ha -f plugins.yaml jfrog/artifactory-ha
 ```
 
+Alternatively, you may be in a situation in which you would like to create a secret in a Helm chart that depends on this chart. In this scenario, the name of the secret is likely dynamically generated via template functions, so passing a statically named secret isn't possible. In this case, the chart supports evaluating strings as templates via the [`tpl`](https://helm.sh/docs/charts_tips_and_tricks/#using-the-tpl-function) function - simply pass the raw string containing the templating language used to name your secret as a value instead by adding the following to your chart's `values.yaml` file:
+```yaml
+artifactory-ha: # Name of the artifactory-ha dependency
+  artifactory:
+    userPluginSecrets:
+      - '{{ template "my-chart.fullname" . }}'
+```
+
 ## Configuration
 The following table lists the configurable parameters of the artifactory chart and their default values.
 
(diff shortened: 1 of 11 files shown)


stable/artifactory-ha/README.md
```diff
@@ -440,6 +440,7 @@ artifactory:
   userPluginSecrets:
     - archive-old-artifacts
     - webhook
+    - '{{ template "my-chart.fullname" . }}'
```
eldada: Can you add a short explanation of this use case. This mix in the example is confusing.

stable/artifactory/README.md
```diff
@@ -353,6 +353,7 @@ artifactory:
   userPluginSecrets:
     - archive-old-artifacts
     - webhook
+    - '{{ template "my-chart.fullname" . }}'
```
eldada: Can you add a short explanation of this use case. This mix in the example is confusing.

stable/artifactory-ha/README.md
```diff
@@ -447,6 +447,14 @@ You can now pass the created `plugins.yaml` file to helm install command to depl
 helm install --name artifactory-ha -f plugins.yaml jfrog/artifactory-ha
 ```
 
+Alternatively, you may be in a situation in which you would like to create a secret in a Helm chart that depends on this chart. In this scenario, the name of the secret is likely dynamically generated via template functions, so passing a statically named secret isn't possible. In this case, the chart supports evaluating strings as templates via the [`tpl`](https://helm.sh/docs/charts_tips_and_tricks/#using-the-tpl-function) function - simply pass the raw string containing the templating language used to name your secret as a value instead by adding the following to your chart's `values.yaml` file:
```
robbie-demuth: Is this more clear, @eldada? I've added the same blob in the `artifactory` README

eldada: Perfect. Thx!

eldada:
@robbie-demuth - please rebase and update versions. Once conflicts are resolved, I'll approve and merge.

rimusz:
nice one 👍

robbie-demuth:
Done, @eldada!

Issue #287, opened 2019-04-03
Configuring userPluginSecrets causes initialization issue due to artifactory_extra_conf not loaded


properpanda:
a bug was introduced with PR#278 (https://github.com/jfrog/charts/pull/278/)  specifically commit:
https://github.com/jfrog/charts/pull/278/commits/cb2316d1985a58999faf126aca67a1dbab6d2572

We have seen behaviour when using userPluginsSecrets such that when configuring a userPluginSecret, a command gets added to the template that creates an /var/opt/jfrog/artifactory/etc folder.

during initialization running entrypoint-artifactory.sh, the script sees that an etc folder exists and DOES NOT add the extra conf files needed in a fresh installation causing faulty installations is undesired effects

from entrypoint-artifactory.sh
```
# In case data dirs are missing or not mounted, need to create them
setupDataDirs () {
    logger "Setting up Artifactory data directories if missing"
    if [ ! -d ${ARTIFACTORY_DATA}/etc ]; then
        mkdir -pv ${ARTIFACTORY_DATA}/etc || errorExit "Failed creating ${ARTIFACTORY_DATA}/etc"

        # Add extra conf files to a newly created etc/ only!
        addExtraConfFiles
    fi
```

We believe that the userPluginsSecrets code should stage plugins from secrets in /tmp/plugins and allow the entrypoint-artifactory.sh to copy them appropriatly.  I can work this bug fix and supply a PR.

properpanda:
I looked at a couple fixes, the most straight forward fix would be to be able to write to the /tmp/plugins directly, but unfortunately that directory on the artifactory-pro:6.9.0 image is drwxr-xr-x root/root.  If we can open the permissions to allow non-root groups to write to it, then the userPluginsSecrets will work with minor adjustments

the alternative solution was to mount a volume on top of /tmp/plugins - either with an emptyDir (and copy files over) or a projected volume that contained all the secrets.  however, the issue was that we lost the prepackaged internalUser.groovy script that exists in /tmp/plugins as part of the artifactory-pro:6.9.0 image.  If people were dependent on this script then it would no longer be there (that's why the initial solution seems simplest and ideal)

jainishshah17:
@properpanda How about mount pluginSecrets to `/tmp/plugin` and then copy them to `/tmp/plugins` using preStartCommand so that Entrypoint can move it over to `ART_HOME/etc/plugins`?

properpanda:
[...]

PR #289, opened 2019-04-03
addressing issue #287 with userPluginSecrets

properpanda:
This PR address issue when using userPluginSecrets described here https://github.com/jfrog/charts/issues/287

the changes prevent the preStartCommand from creating a /var/opt/jfrog/artifactory/etc which creates unintended consequences (as described in the issue reported).

we copy the secrets to /tmp/plugins where the plugins will be picked up and installed.

unfortunately the permissions of /tmp/plugins do not allow you to write to /tmp/plugins as non root which is why we are working around that with an emptyDir volume mount.  making the behaviour such that the only plugins that are loaded via secrets are loaded.

note: you could also changed this to just mount muptiple secrets with a projected volume (https://kubernetes.io/docs/concepts/storage/volumes/#projected) directly to /tmp/plugins avoiding hte manual copy, however this implementation puts us closer to if the /tmp/plugins directory was writable


zendril:
@jainishshah17 @eldada Thoughts on this? It is definitely a bug that was introduced with the userplugins PR and breaks any new/fresh installs.

thanks

jainishshah17:
@properpanda Can you fix this for Artifactory chart as well?

stable/artifactory-ha/templates/artifactory-primary-statefulset.yaml
```diff
@@ -101,8 +101,7 @@ spec:
         - >
           {{- if .Values.artifactory.userPluginSecrets }}
           echo "Copying plugins";
-          mkdir -p {{ .Values.artifactory.persistence.mountPath }}/etc/plugins;
-          cp -Lrf /tmp/plugin/*/* {{ .Values.artifactory.persistence.mountPath }}/etc/plugins/;
```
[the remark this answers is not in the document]
zendril: There is existing code inside the entrypoint.sh that copies all plugins from /tmp/plugins to ART_DATA/etc/plugins.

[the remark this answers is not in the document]
danielezer: oh ok, cool

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -215,6 +214,8 @@ spec:
         - containerPort: {{ .Values.artifactory.internalPort }}
         volumeMounts:
       {{- if .Values.artifactory.userPluginSecrets }}
+        - name: tmp-plugins
```
eldada: I think this mount is not needed as we have the `/tmp/plugins` already created in the image.
By mounting over it, you will mask the `internalUser.groovy` we already have there...

eldada:
Also
```
Error linting charts
-----------------------------------------------------------------------------
 ✔︎ stable/artifactory-ha
 ✖︎ stable/artifactory > Chart version not ok. Needs a version bump!
-----------------------------------------------------------------------------
```

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
properpanda: see all the comments in the related issue: https://github.com/jfrog/charts/issues/287

and the notes initial pr comment:

>  unfortunately the permissions of /tmp/plugins do not allow you to write to /tmp/plugins as non root which is why we are working around that with an emptyDir volume mount. _making the behaviour such that the only plugins that are loaded via secrets are loaded_.

we need /tmp/plugins created from the artifactory image to be writable to appropriately copy into the folder.  If you are using these plugins via secrects , just know/assume that those are the only plugins noted.   I think this behavior is better than the current one of being broken

eldada: Right! My mistake. 
So to cover this, we need to specify this very clearly in the `README.md` and also make a note about `internalUser.groovy`, which is missing that means the auto nginx config update will not work due to its absence.

Anyway - I added the permissions fix to Artifactory Dockerfile, but it will take a few minor releases for it to be public.

stable/artifactory/CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+<<<<<<< HEAD
```
eldada: Fix

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
properpanda: updated

stable/artifactory/CHANGELOG.md
properpanda: fixed

jainishshah17:
Will merge once CI job is done.

PR #294, opened 2019-04-07
Artifactory license

danielezer:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Add support for providing the artifactory license key as a parameter to helm install/upgrade instead of just using a secret name


stable/artifactory-ha/values.yaml
```diff
@@ -235,6 +235,7 @@ artifactory:
   ## ARTIFACTORY_HOME/etc/artifactory.cluster.license and loaded at run time.
   ## The dataKey should be the name of the secret data key created.
   license:
+    licenseKey:
```
eldada: Add a note that only `licenseKey` or `secret` should be used or that `secret` will override `licenseKey` if both are passed.

stable/artifactory/values.yaml
```diff
@@ -175,6 +179,7 @@ artifactory:
   ## ARTIFACTORY_HOME/etc/artifactory.lic and loaded at run time.
   ## The dataKey should be the name of the secret data key created.
   license:
+    licenseKey:
```
eldada: Add a note that only `licenseKey` or `secret` should be used or that `secret` will override `licenseKey` if both are passed.

stable/artifactory-ha/values.yaml
danielezer: @eldada Done

stable/artifactory/values.yaml
danielezer: @eldada Done

danielezer:
@eldada seems like the build is going to timeout for the second time. Is that something we should be worried from?

jainishshah17:
@danielezer Please resolve conflicts.

danielezer:
> @danielezer Please resolve conflicts.

@jainishshah17 Done

PR #315, opened 2019-04-18
updating StatefulSet and Deployment API Version

jainishshah17:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/stable/artifactory-ha/templates/nginx-deployment.yaml
+++ b/stable/artifactory-ha/templates/nginx-deployment.yaml
@@ -1,7 +1,7 @@
 {{- if .Values.nginx.enabled -}}
 {{- $serviceName := include "artifactory-ha.fullname" . -}}
 {{- $servicePort := .Values.artifactory.externalPort -}}
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ template "artifactory-ha.nginx.fullname" . }}
@@ -22,7 +22,9 @@ spec:
     metadata:
       labels:
         app: {{ template "artifactory-ha.name" . }}
+        chart: {{ template "artifactory-ha.chart" . }}
         component: {{ .Values.nginx.name }}
+        heritage: {{ .Release.Service }}
         release: {{ .Release.Name }}
     spec:
       serviceAccountName: {{ template "artifactory-ha.serviceAccountName" . }}
@@ -66,9 +68,7 @@ spec:
               - '/bin/sh'
               - '-c'
               - >
-                {{- if .Values.nginx.customArtifactoryConfigMap }}
-                cp -Lrf /tmp/conf.d/artifactory*.conf /etc/nginx/conf.d/artifactory.conf;
-                {{- else }}
+                {{- if (not .Values.nginx.customArtifactoryConfigMap) }}
                 until [ -f /etc/nginx/conf.d/artifactory.conf ]; do sleep 1 ; done;
                 if ! grep -q 'upstream' /etc/nginx/conf.d/artifactory.conf; then
                 sed -i -e 's,proxy_pass.*http://artifactory.*/artifactory/\(.*\);,proxy_pass       http://{{ $serviceName }}:{{ $servicePort }}/artifactory/\1;,g' \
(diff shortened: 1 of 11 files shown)


stable/artifactory/templates/nginx-deployment.yaml
```diff
@@ -66,9 +68,7 @@ spec:
               - '/bin/sh'
               - '-c'
               - >
-                {{- if .Values.nginx.customArtifactoryConfigMap }}
-                cp -Lrf /tmp/conf.d/artifactory*.conf /etc/nginx/conf.d/artifactory.conf;
```
danielezer: How would this get copied not to /etc/nginx?

jainishshah17: Directly mounting it there

stable/artifactory-ha/templates/nginx-deployment.yaml
```diff
@@ -125,7 +125,8 @@ spec:
         {{- end }}
         {{- if .Values.nginx.customArtifactoryConfigMap }}
         - name: artifactory-nginx-config
-          mountPath: "/tmp/conf.d"
+          mountPath: "/tmp/conf.d/artifactory-ha.conf"
```
danielezer: It’s still mounted to /tmp

stable/artifactory/templates/nginx-deployment.yaml
danielezer: It’s still mounted to /tmp

stable/artifactory/templates/nginx-deployment.yaml
```diff
@@ -125,7 +125,8 @@ spec:
         {{- end }}
         {{- if .Values.nginx.customArtifactoryConfigMap }}
         - name: artifactory-nginx-config
-          mountPath: "/tmp/conf.d"
+          mountPath: "/tmp/conf.d/artifactory.conf"
+          subPath: "artifactory.conf"
```
jainishshah17: @danielezer here

danielezer:
@jainishshah17 it’s pointing to /tmp, not to /etc/nginx. Or am I missing something?

stable/artifactory/templates/nginx-deployment.yaml
jainishshah17: @danielezer My bad. fixing it

stable/artifactory-ha/templates/nginx-deployment.yaml
```diff
@@ -125,7 +125,8 @@ spec:
         {{- end }}
         {{- if .Values.nginx.customArtifactoryConfigMap }}
         - name: artifactory-nginx-config
-          mountPath: "/tmp/conf.d"
+          mountPath: "/etc/nginx/conf.d/artifactory-ha.conf"
+          subPath: "artifactory-ha.conf"
```
danielezer: Shouldn’t this be configurable? Or at least documented? I mean, the fact that you have to have a key called artifactory-ha.conf here and artifactory.conf in the artifactory chart

jainishshah17: @danielezer  https://github.com/jfrog/charts/blob/master/stable/artifactory-ha/ReverseProxyConfiguration.md#steps-to-use-static-configuration-for-reverse-proxy-in-nginx

danielezer: I see that but I think it’s worth mentioning that the data key has to be called by that specific name. I’m not I would understand that based in the current documentation. But that’s just me 😅

jainishshah17: @danielezer I do agree. Do you want me to make it configurable or just add note in the readme?

danielezer: Mmm whatever you think 🙂
I think both are good

jainishshah17:
> I like this change. Good clean up.
> Does it require to update minimal K8s version supported?

It should work with all version.

stable/artifactory-ha/templates/nginx-deployment.yaml
danielezer: @jainishshah17 what about this change? what do you think?

jainishshah17: @danielezer Done. PTAL.

PR #319, opened 2019-04-25
Added support for cacheProviderDir

claasd:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

This PR allows to set the `cacheProviderDir`. As a result, one can use persistent volume claims together with Azure BLOB Storage when settings `cacheProviderDir` and `mountPath` to the same directory.

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.12.19] - Apr 28, 2019
+* Added support for `cacheProviderDir`
+
 ## [0.12.18] - Apr 18, 2019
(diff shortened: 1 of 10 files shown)


eldada:
@claasd - Thanks for the contribution. Can you also add this to the `artifactory-ha` chart?

claasd:
@eldada I added the changes to artifactory-ha. 
However, the Azure Blob Storage configuration for HA does not contain a cache-fs configuration in the `artifactory-binstore` file (unlike google or S3). We are not using artifactory-ha, therefore I cannot say if there should be this section.

eldada:
They should be similar. Can you make them the same?

claasd:
Hi, I added the cache-fs configuration to Azure-Blob for Artifactory-HA.

sanjeevkarani:
@eldada 
The support for 'artifactory.persistence.cacheProviderDir' is limited to cloud storage providers.
It is also required for 'file-system' persistence of on-premise k8s platform

PR #330, opened 2019-05-10
Adding custom volume mounts

kvonbredow:
Signed-off-by: Kyle von Bredow <kylevonbredow@gmail.com>

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

This PR adds support for custom volumes to be mounted into the mission control container.  This will allow the custom init containers to modify the filesystem before mission control runs.  For example, we can add an enterprise CA into the trusted root store, so that mission control can sync to an internally hosted git repository.

### DIFF
--- a/stable/mission-control/CHANGELOG.md
+++ b/stable/mission-control/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Mission-Control Chart Changelog
 All changes to this chart will be documented in this file.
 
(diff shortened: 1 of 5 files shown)


kvonbredow:
Thanks for getting back to me so quickly! I've changed the implementation to match that of the other charts. Let me know if everything looks right

eldada:
@amithins tested PR locally and approved changes are ok. Merging.

PR #332, opened 2019-05-12
Allow override of busybox in rabbitmq-ha

### LINKED ISSUE #331: Use newer rabbitmq-ha >= 1.13.1 in xray

---

BUG REPORT

Helm 2.12.3
Kubernetes 1.10.11

**Which chart**:
xray

**What happened**:
As we need to change the image registry, it could not pull image busybox in rabbitmq-ha. The image name is hardcoded in this old version.

Please use a newer version of rabbitmq-ha.

jonasrutishauser:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Allow override of busybox in rabbitmq-ha

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #331 

I diffed to outcome using the `test-values.yaml` and i think there is no relevant change in the generated k8s objects.
Therefore it will support the two requirements asked in the issue.

### DIFF
--- a/stable/xray/CHANGELOG.md
+++ b/stable/xray/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Xray Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.12.5] - May 12, 2019
+* Updated rabbitmq-ha chart version to 1.13.1
+
 ## [0.12.4] - Apr 15, 2019
 * Simplify handling connection strings setup in `xray_config.yaml` to better support ampersand in external connection strings
 * **IMPORTANT:** If using an external connection string for PostgreSQL or MongoDB, **do not escape** the ampersand with `\`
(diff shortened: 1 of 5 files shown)


eldada:
@jonasrutishauser - Thanks for the PR!
1. Does this chart version support upgrading from existing version?
2. Any reason not to go with a higher chart version?

jonasrutishauser:
I choose the version as our needed fix is there included.

I just made a diff with the newest version 1.26.0 and i think we can even use that version.
It will reference rabbitmq 3.7.12-alpine but if needed it can be changed to 3.7-alpine.
I will change the PR to use the newest version.

I think there shouldn't be any problem upgrading from a previous version.

I tested it in our company (with the 1.26.0 version) and the only problem was that i had to increase the `initialDelaySeconds` in the livenessProbe.

PR #336, opened 2019-05-16
Support 'artifactory.persistence.cacheProviderDir' for on-premise cluster

sanjeevkarani:
…premise cluster

#### PR Checklist
- [ ] Chart Version bumped

### DIFF
--- a/stable/artifactory-ha/templates/artifactory-binarystore.yaml
+++ b/stable/artifactory-ha/templates/artifactory-binarystore.yaml
@@ -24,6 +24,11 @@ data:
             </provider>
         </chain>
         
+        <provider id="cache-fs" type="cache-fs">
+            <maxCacheSize>{{ .Values.artifactory.persistence.maxCacheSize }}</maxCacheSize>
+            <cacheProviderDir>{{ .Values.artifactory.persistence.cacheProviderDir }}</cacheProviderDir>
+        </provider>
+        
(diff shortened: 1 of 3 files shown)


stable/artifactory-ha/templates/artifactory-binarystore.yaml
```diff
@@ -24,6 +24,11 @@ data:
             </provider>
         </chain>
         
+        <provider id="cache-fs" type="cache-fs">
```
danielezer: I think this should be wrapped with an if statement. However, if you feel that this is always needed, there should be a default value in values.yaml. In the current state, this can cause issues when you don't pass these values.

stable/artifactory-ha/templates/artifactory-binarystore.yaml
```diff
@@ -55,6 +60,11 @@ data:
             </provider>
         </chain>
 
+        <provider id="cache-fs" type="cache-fs">
```
danielezer: Same goes here

danielezer:
@sanjeevkarani ^^

sanjeevkarani:
Hi Daniel,

I see your point.
But, I see that the default values are already defined in 'values.yaml'
file.
Am I missing something?

artifactory:
persistence: maxCacheSize: 50000000000 cacheProviderDir: cache

danielezer:
@sanjeevkarani oh sorry. I missed that. 
Thanks.

PR #337, opened 2019-05-17
update kind and ct to latest versions

rimusz:


.travis.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         - echo "Run charts-testing install test!"
         # ping stdout every 9 minutes or Travis kills the build
```
eldada: Change `9` to `20`?

rimusz: it might help to keep kind longer up

PR #368, opened 2019-06-17
[artifactory] nginx config maps

### LINKED ISSUE #308: nginx keeps crashing with the following error message: nginx: [emerg] host not found in upstream "artifactory
artifactory version: artifactory-pro:6.9.0 

chart: artifactory HA 

**What happened**: 
image002 (2)

**What you expected to happen**: nginx to identify the changes and run

**How to reproduce it** (as minimally and precisely as possible): kubernetes and baremetal with Metallb as a l2 loadbalancer

### LINKED ISSUE #312: use a configmap for nginx configuration
NO
---

FEATURE REQUEST

ALL

**Which chart**:
artifactory and artifactory-ha

**What happened**:
When installing the artifactory and artifactory-ha charts, nginx will fail to start one time because there's a sort of race conditions between the nginx startup and the postStart command.

nginx to start properly without restarting

helm install jfrog/artifactory

I think that if we start using a configmap for the nginx configuration and mount it directly onto /etc/nginx/nginx.conf it would remove the need to use sed and copy to generate the nginx configuration, thus removing the need for the postStart container lifecycle.

### LINKED ISSUE #352: providing customArtifactoryConfigMap will result in nginx crashloop
custom nginx config can be provided via customArtifactoryConfigMap option. However, enable this option right now will result in crashed nginx pod. 

When describing the pod, the following error can be found:

```
Warning  FailedPostStartHook     1m (x3 over 1m)  kubelet, gke-nutanix-nginx-default-pool-91516bba-qwfw  Exec lifecycle hook ([/bin/sh -c if [ -f /tmp/replicator-nginx.conf ]; then cp -fv /tmp/replicator-nginx.conf /etc/nginx/conf.d/replicator-nginx.conf; fi; if [ -f /tmp/ssl/*.crt ]; then rm -rf /var/opt/jfrog/nginx/ssl/example.*; cp -fv /tmp/ssl/* /var/opt/jfrog/nginx/ssl; fi; sleep 5; nginx -s reload; touch /var/log/nginx/conf.done
])
```
And from the container logs:

`nginx: [emerg] host not found in upstream "artifactory" in /etc/nginx/conf.d/artifactory.conf:31`

It looks like the cause of the issue is that:

1. We expect artifactory.conf that passed with config map to be mounted to /etc/nginx/conf.d/artifactory.conf (which it do). But the entrypoint script in the nginx container took the template under /var/opt/jfrog/nginx/artifactory.conf and replace the one we passed in when the container starts.

2. when 'customArtifactoryConfigMap' is not false, the following script is skipped during poststart life cycle, which didn't point to the correct service if you are installing with release name other than 'artifactory'. This script is expected to be skipped but, since what's under /etc/nginx/conf.d/ didn't get replaced by the custom artifact.conf passed in, it causes the "host not found in upstream" error

sed -i -e 's,proxy_pass.*http://artifactory.*/artifactory/\(.*\);,proxy_pass       http://{{ $serviceName }}:{{ $servicePort }}/artifactory/\1;

A possible fix is to modify nginx template to mount customArtifactoryConfigMap to a tmp folder first and use script to copy it to the right location just like the ssl secret. 

```
lifecycle:
          postStart:
            exec:
              command:
              - '/bin/sh'
              - '-c'
              - >
                {{- if (not .Values.nginx.customArtifactoryConfigMap) }}
                until [ -f /etc/nginx/conf.d/artifactory.conf ]; do sleep 1 ; done;
                if ! grep -q 'upstream' /etc/nginx/conf.d/artifactory.conf; then
                sed -i -e 's,proxy_pass.*http://artifactory.*/artifactory/\(.*\);,proxy_pass       http://{{ $serviceName }}:{{ $servicePort }}/artifactory/\1;,g' \
                    -e 's,server_name .*,server_name ~(?<repo>.+)\\.{{ $serviceName }} {{ $serviceName }};,g' \
                    /etc/nginx/conf.d/artifactory.conf;
                fi;
                {{- end }}
                **cp -fv /tmp/nginx/artifactory.conf /etc/nginx/conf.d/artifactory.conf;**
                if [ -f /tmp/replicator-nginx.conf ]; then
                cp -fv /tmp/replicator-nginx.conf /etc/nginx/conf.d/replicator-nginx.conf;
                fi;
                if [ -f /tmp/ssl/*.crt ]; then
                rm -rf /var/opt/jfrog/nginx/ssl/example.*;
                cp -fv /tmp/ssl/* /var/opt/jfrog/nginx/ssl;
                fi;
                sleep 5; nginx -s reload; touch /var/log/nginx/conf.done
```

```
{{- if .Values.nginx.customArtifactoryConfigMap }}
        - name: artifactory-nginx-config
          mountPath: "/tmp/nginx/artifactory.conf"
          subPath: "artifactory.conf"
{{- end }}
```

danielezer:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Use configmaps for nginx configuration instead of using sed to replace text in the config files. This will solve the nginx restart issues we have been experiencing from different customers by removing the race condition for replacing the text in the nginx configuration.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 
fixes #312, fixes #352, fixes #308,


stable/artifactory/templates/nginx-artifactory-conf.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+      if ($http_x_forwarded_proto = '') {
+        set $http_x_forwarded_proto  $scheme;
+      }
+      ## Application specific logs
+      ## access_log /var/log/nginx/artifactory-access.log timing;
+      ## error_log /var/log/nginx/artifactory-error.log;
+      rewrite ^/$ /artifactory/webapp/ redirect;
+      rewrite ^/artifactory/?(/webapp)?$ /artifactory/webapp/ redirect;
+      rewrite ^/(v1|v2)/(.*) /artifactory/api/docker/$repo/$1/$2;
```
matankatz: To support Docker repositories to work without sub domains, this rewrite rule should be valid if repo variable is not null.
`if ( $repo != "" ){
    rewrite ^/(v1|v2)/(.*) /artifactory/api/docker/$repo/$1/$2;
  }`

danielezer: @matankatz Fixed, thanks

qwertyberty:
Fails to start nginx container when using an existing volume claim, with error:
```bash
PostStartHookError: command '/bin/sh -c ln -s /dev/stdout /var/log/nginx/access.log; ln -s /dev/stderr /var/log/nginx/error.log;
' exited with 1: ln: /var/log/nginx/access.log: File exists
ln: /var/log/nginx/error.log: File exists
```

danielezer:
@qwertyberty Is this related to this pull request? This code is not merged to master yet. You should create a separate issue and we will address it.

PR #378, opened 2019-06-20
add kubeval to lint, increase install timeout

rimusz:
Add [kubeval](https://kubeval.instrumenta.dev/) to validate k8s manifests, increase timeout for `ct install`


test/lint-charts.sh
```diff
@@ -24,11 +38,15 @@ get_changed_charts() {
...
+    echo " Checking CHANGELOG!"
+    echo " Charts to be processed: ${changed_charts[*]}"
+    echo "------------------------------------------------------------------------------------------------------------------------"
     for chart_name in ${changed_charts[*]} ; do
-        echo "Checking CHANGELOG for chart ${chart_name}"
+        echo "==> Checking CHANGELOG for chart ${chart_name}"
+        echo "------------------------------------------------------------------------------------------------------------------------"
         local chart_version
-        chart_version=$(grep "version:" "${REPO_ROOT}/${chart_name}/Chart.yaml")
+        chart_version=$(grep "version:" "${REPO_ROOT}/${chart_name}/Chart.yaml" | cut -d":" -f 2 | cut -d" " -f 2)
```
danielezer: You don't need the double "cut", only one, based on the " " is enough

PR #382, opened 2019-06-23
Add values files for small, medium and large installations

### LINKED ISSUE #9: [Proposal] Provide different values for small (default), medium & large deployments
**What**
Ship 3 flavors of values.yaml so that a user (dev, SML customer) can pick the flavor of the file that's more relevant and is only exposed to a very minimum set of properties that they have to modify, versus the entire set. There is no need for the user to touch the main `values.yaml` file and instead they should only be modifying the relevant file that includes a very minimum set. And then pass the modular file via  -f custom-values-file option.
 

**Why**
Having a single `values.yaml` is too much especially for customers who are new to helm. Relevant details such compute, secret names are missed because they are hidden among 100s of key-value pairs. Plus these values are not optimized for their use case. Example: For a customer that runs 300 concurrent builds with 1GB payload, the default compute values are not enough. Or in case of Xray, the memory assigned to MongoDB should be atleast 6GB for staging and maybe 12GB for production. This would enhance the UX for our customers and even reduce the readme size.

danielezer:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Add values files for small, medium and large installations

fixes #9

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [0.14.1] - Jun 23, 2019
+* Add values files for small, medium and large installations
+
 ## [0.14.0] - Jun 20, 2019
 * Use ConfigMaps for nginx configuration and remove nginx postStart command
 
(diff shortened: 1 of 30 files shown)


danielezer:
@rimusz Done

danielezer:
@nimerb The current value for Xray is set to `--wiredTigerCacheSizeGB=3`

PR #403, opened 2019-06-27
Add the option for changing nginx config using values.yaml and remove…

danielezer:
… outdated reverse proxy documentation for artifactory-ha

#### PR Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] CHANGELOG.md updated
- [x] Variables and other changes are documented in the README.md

**What this PR does / why we need it**:
Add the option to provide nginx configuration with a values file

PR #408, opened 2019-07-02
[Artifactory] Update nginx config to detect docker path or subdomain

btashton:
Currently the nginx config only rewrites docker api requests for
subdomain style repositories.  This detects if a path based repo
is being used instead and properly rewrites those requests.

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Currently path based docker repositories are not supported by the default nginx configuration.  This change supports both path and subdomain configurations.

I do not have a configuration to verify the subdomain case, but it should be handled by the null repo check which now has a break statement to prevent further rewriting of the docker api request.

### DIFF
--- a/stable/artifactory/CHANGELOG.md
+++ b/stable/artifactory/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [7.16.2] - Jul 1, 2019
+* Allow path based docker repos in the nginx config
+
 ## [7.16.1] - Jul 1, 2019
 * Updated Artifactory version to 6.11.1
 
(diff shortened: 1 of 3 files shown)


rimusz:
@btashton can you add support of artifactory-ha as well, thanks

btashton:
@rimusz I think there is an outstanding PR #403 that is moving around the nginx config so it might be best to make that change at the same time?

eldada:
> @rimusz I think there is an outstanding PR #403 that is moving around the nginx config so it might be best to make that change at the same time?

Yes. I asked @danielezer to add the changes here to his PR (#403).

danielezer:
@btashton Added these changes to my PR #403

btashton:
Closing as changes are now part of #403

PR #414, opened 2019-07-11
[artifactory-ha] Add ingress.hosts to the Nginx server_name directive when ingress is enabled

danielezer:
… when ingress is enabled

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Add the ingress.hosts to the nginx server_name directive. This will make sure that the ingress hostnames are able to resolve the repository name from the hostname by the Docker repository sub domain method. This will help users with setting up Docker repository sub domain access without any need to provide a custom values.yaml


danielezer:
@rimusz :)

PR #426, opened 2019-07-24
[artifactory] Simplify nginx setup

eldada:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Now that nginx config is managed in `values.yaml`, we can simplify init setup of the nginx pod.

### DIFF
--- a/stable/artifactory/templates/nginx-deployment.yaml
+++ b/stable/artifactory/templates/nginx-deployment.yaml
@@ -37,37 +37,18 @@ spec:
       - name: {{ .Values.imagePullSecrets }}
     {{- end }}
       initContainers:
-    {{- if .Values.nginx.persistence.enabled }}
-      - name: "remove-lost-found"
+      - name: "setup"
         image: "{{ .Values.initContainerImage }}"
         imagePullPolicy: {{ .Values.nginx.image.pullPolicy }}
         command:
         - '/bin/sh'
         - '-c'
-        - 'rm -rfv {{ .Values.nginx.persistence.mountPath }}/lost+found'
+        - >
+          [ -d {{ .Values.nginx.persistence.mountPath }}/lost+found ] && rm -rfv {{ .Values.nginx.persistence.mountPath }}/lost+found;
+          mkdir -p {{ .Values.nginx.persistence.mountPath }}/logs;
         volumeMounts:
         - mountPath: {{ .Values.nginx.persistence.mountPath | quote }}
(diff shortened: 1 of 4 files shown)


stable/artifactory/templates/nginx-deployment.yaml
```diff
@@ -37,37 +37,18 @@ spec:
...
-      - name: "remove-lost-found"
+      - name: "setup"
         image: "{{ .Values.initContainerImage }}"
         imagePullPolicy: {{ .Values.nginx.image.pullPolicy }}
         command:
         - '/bin/sh'
         - '-c'
-        - 'rm -rfv {{ .Values.nginx.persistence.mountPath }}/lost+found'
+        - >
+          [ -d {{ .Values.nginx.persistence.mountPath }}/lost+found ] && rm -rfv {{ .Values.nginx.persistence.mountPath }}/lost+found;
```
danielezer: @eldada is it really necessary to check I
if the directory exists? mew -rf will be good even when the directory doesn’t exist

stable/artifactory/templates/nginx-deployment.yaml
```diff
@@ -37,37 +37,18 @@ spec:
...
-        volumeMounts:
-          - mountPath: {{ .Values.nginx.persistence.mountPath | quote }}
-            name: nginx-volume
-      - name: "wait-for-artifactory"
-        image: "{{ .Values.initContainerImage }}"
-        command:
-        - 'sh'
-        - '-c'
-        - >
-          until nc -z -w 2 {{ $serviceName }} {{ $servicePort }} && echo artifactory ok; do
```
danielezer: What about the wait-for-artifactory? 
Wouldn’t this cause a lot of nginx restarts?

rimusz: I should keep the `wait-for-artifactory`

stable/artifactory/templates/nginx-deployment.yaml
eldada: Agree. Removed.

stable/artifactory/templates/nginx-deployment.yaml
eldada: It was restarting anyway... the wait-for-artifactroy was not doing a good job anyway. I'd rather not use a faulty init container and simplify the setup. 
It's better the pos restarts on liveness than on faulty config.

PR #437, opened 2019-08-05
[artifactory] Improve handling of Access bootstrap.creds

eldada:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

* Do not mount `access/etc/bootstrap.creds` unless user specifies a custom password or secret (Access already generates a random password if not provided one)
* If custom `bootstrap.creds` is provided (using keys or custom secret), prepare it with an init container so the temp file does not persist


stable/artifactory/templates/access-bootstrap-creds.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if not (and .Values.artifactory.accessAdmin.secret .Values.artifactory.accessAdmin.dataKey) }}
+{{- if .Values.artifactory.accessAdmin.password }}
```
danielezer: @eldada I actually think it’s good that we use the same method for when you provide a secret and when you don’t. This way you always know where it is stored instead of it being sometimes on disk and sometimes with a secret. WDYT?

eldada:
Tested locally. `make gke -- --charts stable/artifactory` passed.

PR #513, opened 2019-10-24
Artifactory New postgresql sub chart version

### LINKED ISSUE #286: [Proposal] Upgrade postgresql chart dependency (for artifactory and artifactory-ha) to get securityContext and pgsql 10.x
No
---

FEATURE REQUEST

**Which chart**:
artifactory and artifactory-ha

**What happened**:
We currently pull in a really old version of the postgresql chart which is 9.x postgresql install.

This is an issue because these older versions of the chart do not support securityContext which  then breaks many installs where pods are not allowed to be run as root.

Upgrade to a newer version of the postgresql chart that pulls in postgresql 10.x and supports securityContext.

According to https://www.jfrog.com/confluence/display/RTF/Configuring+the+Database artifactory will support Postgresql >= 9.2. 

Part of the real reason for asking to upgrade the chart is that the old version of the postgresql chart does not allow for securityContext, so I have to manually override the template to include it given that we cannot run our pods as root.

eldada:
# DO NOT MERGE #

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Upgrade PostgrSQL sub-chart version to support statefulset DB and K8s 1.16.x (deprecated `apiVersion`).

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #286 

Follow upgrade notes in `CHANGELOG.md`.


stable/artifactory-ha/requirements.yaml
```diff
@@ -1,5 +1,5 @@
 dependencies:
-- name: postgresql
-  version: 0.9.5
-  repository: https://kubernetes-charts.storage.googleapis.com/
-  condition: postgresql.enabled
+  - name: postgresql
+    version: 6.5.3
+    repository: https://kubernetes-charts.storage.googleapis.com/
+    condition: postgresql.enabled
```
vinayagg: Is it possible to change this to .Values.postgresql.enabled ?
The reason being, this change would make it compatible to be embedded into other charts (like JCR).

PR #517, opened 2019-11-03
Make sure the artifactory pod exits when one of the pre-start stages …

### LINKED ISSUE #511: Pod successfully starts even if `copyOnEveryStartup` commands fail
No
---

BUG REPORT

Latest

**Which chart**:
artifactory-ha & artifactory

**What happened**:
I've provided a command within the `copyOnEveryStartup`. It failed. Nonetheless, the pod was started successfully with no indication for the failure.

I would expect the pod to fail, or at least get an indication that exit code wasn't 0.

Within the `copyOnEveryStartup` section try to copy something into a directory that doesn't exist.

danielezer:
…fail

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

When one of the pre-start stages (e.g. copyOnEveryStartup) fails, the artifactory pod needs to fail and exit. This fix will cause it to do that.

fixes #511 

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.0.1] - Nov 3, 2019
+* Make sure the artifactory pod exits when one of the pre-start stages fail
+
 ## [1.0.0] - Oct 27, 2019
 **IMPORTANT - BREAKING CHANGES!**<br>
 **DOWNTIME MIGHT BE REQUIRED FOR AN UPGRADE!**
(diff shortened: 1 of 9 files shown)


danielezer:
@rimusz I'm merging this PR but I don't like the CI error:
```
Error linting charts: Error building dependencies for chart 'artifactory-ha => (version: "1.0.1", path: "stable/artifactory-ha")': Error waiting for process: exit status 1
```
From looking in the chart-testing code, it seems like this error message came from:
```
func (h Helm) BuildDependencies(chart string) error {
	return h.exec.RunProcess("helm", "dependency", "build", chart)
}
```

But when I run CI locally, or even when I run: `helm dep build <chart>` it works as expected. Do you have an idea to what might be going on?

PR #528, opened 2019-11-07
Add support for any type of environment variable by using `extraEnvir…

### LINKED ISSUE #519: Allow env to be sourced from secrets

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): FEATURE REQUEST

**Version of Helm and Kubernetes**: 2.9 Helm, 1.15.3 Kubernetes

**What happened**: The helm chart allows you to pass extra env vars, but they are just static values for the most part. 

**What you expected to happen**: I would like to have secrets mounted as env vars in the artifactory primary and member nodes.  This is possible like so: 
 https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

**Anything else we need to know**: I want to do this as we are working with you guys on getting art-ha and access-federation working better in a k8s environment. One of the things we need to support is additional logback appenders (like the logstash appender to send things to kafka and EFK stack). I have most of that working except I need to get the password into the logback configuration, which it is able to read from env vars. 

So you can do something like this in the logback.xml:
`<producerConfig>ssl.keystore.password=${LOGA_KEYSTORE_SECRET}</producerConfig>` 

Need something like extraEnvironmentVariables but allows the value type to be set to not just 'value" but could also be something like 'valueFrom'.

danielezer:
…onmentVariables` as-is

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Add support for any time of environment variable by using `extraEnvironmentVariables` as-is, thus allowing the usage of environment variables from secrets (among others)

fixes #519 

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -2,6 +2,9 @@
 All changes to this chart will be documented in this file.
 
 ## [1.1.2] - Nov 5, 2019
+* Add support for any time of environment variable by using `extraEnvironmentVariables` as-is
+
+## [1.1.2] - Nov 5, 2019
 * Add support for the aws-s3-v3 filestore, which adds support for pod IAM roles
 
 ## [1.1.1] - Nov 4, 2019
--- a/stable/artifactory-ha/Chart.yaml
+++ b/stable/artifactory-ha/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: artifactory-ha
 home: https://www.jfrog.com/artifactory/
-version: 1.1.2
+version: 1.1.3
 appVersion: 6.14.0
 description: Universal Repository Manager supporting all major packaging formats,
   build tools and CI servers.
--- a/stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
+++ b/stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
@@ -280,10 +280,9 @@ spec:
         - name: HA_BACKUP_DIR
           value: "{{ .Values.artifactory.persistence.nfs.backupDir }}"
       {{- end }}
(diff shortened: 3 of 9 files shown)


zendril:
Yeah, this LGTM. Thanks @danielezer

kpirateone:
This update changed the behavior of templated values.

using a template to extract the first of the list of ingress hosts and setting an environment variable used to work with the old implementation

`- name: PRIMARY_INGRESS_HOST
value: "{{- range $index, $host := .Values.ingress.hosts }}{{ if eq $index 0 }}{{ $host }}{{- end }}{{- end }}" `

we now see an error with the new implementation:

> Error: render error in "artifactory-ha/templates/artifactory-primary-statefulset.yaml": template: artifactory-ha/templates/artifactory-primary-statefulset.yaml:323:3: executing "artifactory-ha/templates/artifactory-primary-statefulset.yaml" at <tpl (toYaml .) $>: error calling tpl: Error during tpl function execution for "... redacted ...": parse error in "artifactory-ha/templates/artifactory-primary-statefulset.yaml": template: artifactory-ha/templates/artifactory-primary-statefulset.yaml:7: unexpected unclosed action in command

it seems like complex nesting of templateing breaks the toYaml

removing the the if condition works (with slightly different intended outcome):

`value: "{{- range $index, $host := .Values.ingress.hosts }}{{ $host }}{{- end }}" 
`

so, i dont know if its a bug with toYaml?

(also, i know i can get the value using index: `value: "{{ index .Values.ingress.hosts 0 }}"`, but above does demonstrate the issue)

danielezer:
Thanks @kpirateone, I tried it out as well, and it really does seem like a `toYaml` bug. However, I think this is a pretty rare use case to have such complicated nested expressions. @eldada WDYT?

PR #531, opened 2019-11-08
Details on masterKey

gliptak:
https://stackoverflow.com/a/55113540/304690

### DIFF
--- a/stable/artifactory/CHANGELOG.md
+++ b/stable/artifactory/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.1.7] - Nov 9, 2019
+* Additional documentation for masterKey
+
 ## [8.1.6] - Nov 10, 2019
 * Update PostgreSQL chart version to 7.0.1
 * Use formal PostgreSQL configuration format
--- a/stable/artifactory/Chart.yaml
+++ b/stable/artifactory/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: artifactory
 home: https://www.jfrog.com/artifactory/
-version: 8.1.6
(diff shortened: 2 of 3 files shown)


eldada:
Hi @gliptak . Thx for the PR. The CI failed as you did not update chart version. Please bump chart version in `Chart.yaml` and add entry in `CHANGELOG.md` explaining the change.
BTW - your StackOverflow link is Artifactory HA. This PR is distribution... is this intentional?

gliptak:
@eldada thank you for the pointers. Updated. I'm not sure of that difference between HA and distribution ...

eldada:
@gliptak - you have a mix up.
The changes in `stable/distribution/README.md` are for the *JFrog Distribution Helm Chart*. 
The changes in `stable/artifactory/CHANGELOG.md` and `stable/artifactory/Chart.yaml` are for the *JFrog Artifactory Helm Chart*.
These are two different charts to different products.
I think you mean to add the changes to the README of [Artifactory](https://github.com/jfrog/charts/tree/master/stable/artifactory) and [Artifactory HA](https://github.com/jfrog/charts/tree/master/stable/artifactory-ha) charts?

gliptak:
@eldada updated

eldada:
I did some digging to get the actual format of the master key string.
The official definition is: `A 128-Bit key size (hexadecimal encoded)`. Not just a 16 characters log string.
It can be generated with 
```
openssl rand -hex 16
```
 
So can you please update the description text to
```
Artifactory masterkey. A 128-Bit key size (hexadecimal encoded) string (32 hex characters). Can be generated with `openssl rand -hex 16`.
NOTE: This key can be generated only once and cannot be updated once created.
```

How does this sound?

gliptak:
@eldada thank you for looking into this. I guess one cannot just type in a masterKey ... I updated as per above

PR #539, opened 2019-11-11
add service annotations options to mission control chart. 

### LINKED ISSUE #538: [mission-control] provide annotations options to service so that user can use it to allocate private LB
Some cloud provider allows auto-allocation of LB resources with private IP when certain annotations were given. However, they often do not work with the default ingress controller. However, unlike the artifactory chart, if we choose to expose mission control over the LoadBalancer service type, we do not have the option to provide annotations currently. 

Examples on GKE:

annotations:
      cloud.google.com/load-balancer-type: "Internal"

Propose to add such options to the charts.

See pull request below.

ypan887:
#### PR Checklist
- [x ] Chart Version bumped
- [x ] CHANGELOG.md updated
- [x ] Variables and other changes are documented in the README.md

added service annotations options to mission control chart. So that user can provide annotations to mission control service. This is important if user wants to expose mission control with LB service type but want that LB to have private IP. This can not be done with ingress annotations, because for many provider, the default ingress controller does not work with private LB. 

fixes #538


danielezer:
@ypan887 Please resolve the conflicts

PR #562, opened 2019-11-21
Make the Xray probes customisable

### LINKED ISSUE #542: "livenessProbe" of Xray microservices should be larger during upgrades
---
**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): Bug report.

**Version of Helm and Kubernetes**: Not relevant.

**Which chart**: Xray.

**What happened**: During an upgrade of Xray fro, 2.6.X to 2.7.X or larger, there is a large MongoDB migration, "V0060_RootFiles_Indicator_Added_Up". In large environments, this migration can take several minutes to complete. As there is a "livenessProbe" and the application doesn't respond yet, the migration fails in the middle. 
As the first migration failed due to pod restart, it leaves a lock on the MongoDB database and the application cannot resume after the restart.

**What you expected to happen**: I suggest (not sure if possible) that during upgrades, the "livenessProbe" should be 1 hour.

**How to reproduce it** (as minimally and precisely as possible): Setup Xray 2.6.X with large enviornment with over 1 million of root files, so the "V0060_RootFiles_Indicator_Added_Up" should take some time to complete, then upgrade to 2.7.X or later.

To workaround this once it happened we had to remove the "livenessProbe" from the Xray microservices for the migration, and to remove the lock from [MongoDB](https://www.jfrog.com/confluence/display/XRAY/Troubleshooting#Troubleshooting-Xraydatabasemigrationfailedandneedstoberetried).

danielezer:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Make the Xray probes customisable 

fixes #542 

@arielkv this PR will make the probes customisable so that you can change the probe to your needs. For example, before a big upgrade, you can change the `initialDelayPeriosSeconds` param to 1 hour. This should be enough to fix your issue, since we can't really change the probe only on upgrade because not all upgrade are as big as you describe and having such a long initial delay would be problematic.

### DIFF
--- a/stable/xray/CHANGELOG.md
+++ b/stable/xray/CHANGELOG.md
@@ -2,6 +2,9 @@
 All changes to this chart will be documented in this file.
 
 ## [1.2.7] - Nov 21, 2019
+* Make the Xray probes customisable 
+
+## [1.2.7] - Nov 21, 2019
 * Prevent probes failing on 403 (Forbidden) - fixes
 
 ## [1.2.6] - Nov 20, 2019
--- a/stable/xray/CHANGELOG.md
+++ b/stable/xray/CHANGELOG.md
@@ -1,7 +1,7 @@
 # JFrog Xray Chart Changelog
 All changes to this chart will be documented in this file.
 
-## [1.2.7] - Nov 21, 2019
+## [1.2.8] - Nov 21, 2019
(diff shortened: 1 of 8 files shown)


eldada:
Fix failing CI

PR #582, opened 2019-12-10
Adds the ability to mount a single custom ca chain into the xray server and artifactory pods

evannorstrand:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

What this PR does / why we need it:
This PR adds the ability to include an custom CA certificate. This is needed for enclaved environments, as well as corporate environments that use their own CA certs.

Which issue this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close that issue when PR gets merged): fixes #

Special notes for your reviewer:
Tested on our production environment, it works.

### DIFF
--- a/stable/artifactory/templates/artifactory-statefulset.yaml
+++ b/stable/artifactory/templates/artifactory-statefulset.yaml
@@ -276,6 +276,11 @@ spec:
         - containerPort: {{ .Values.artifactory.javaOpts.jmx.port }}
         {{- end }}
         volumeMounts:
+      {{- if .Values.customCA.enabled }}
+        - name: custom-ca
+          mountPath: "/etc/ssl/certs/custom-ca.crt"
+          subPath: "custom-ca.crt"
+      {{- end}}
       {{- if .Values.artifactory.userPluginSecrets }}
         - name: tmp-plugins
           mountPath: "/tmp/plugins/"
@@ -391,6 +396,11 @@ spec:
 {{ toYaml . | indent 8 }}
     {{- end }}
       volumes:
+      {{- if .Values.customCA.enabled}}
+      - name: custom-ca
+        configMap:
+          name: {{ .Values.customCA.name }}
+      {{- end }}
       - name: binarystore-xml
         secret:
           {{- if .Values.artifactory.persistence.customBinarystoreXmlSecret }}
--- a/stable/xray/templates/xray-server-statefulset.yaml
+++ b/stable/xray/templates/xray-server-statefulset.yaml
@@ -147,6 +147,11 @@ spec:
         volumeMounts:
         - name: data-volume
           mountPath: "{{ .Values.common.xrayConfigPath }}"
+        {{- if .Values.customCA.enabled }}
+        - name: custom-ca
+          mountPath: "/etc/ssl/certs/custom-ca.crt"
+          subPath: "custom-ca.crt"
+        {{- end}}
         securityContext:
           allowPrivilegeEscalation: false
         resources:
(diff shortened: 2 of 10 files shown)


evannorstrand:
This replaces my previous PR, I have added the fixes to artifactory as well.

evannorstrand:

stable/xray/templates/xray-server-statefulset.yaml
```diff
@@ -147,6 +147,11 @@ spec:
         volumeMounts:
         - name: data-volume
           mountPath: "{{ .Values.common.xrayConfigPath }}"
+        {{- if .Values.customCA.enabled }}
+        - name: custom-ca
+          mountPath: "/etc/ssl/certs/custom-ca.crt"
+          subPath: "custom-ca.crt"
```
danielezer: This means that the key has to be set to custom-ca.crt
If that’s how you intend it to be, it should be well documented

stable/artifactory/templates/artifactory-statefulset.yaml
```diff
@@ -276,6 +276,11 @@ spec:
         - containerPort: {{ .Values.artifactory.javaOpts.jmx.port }}
         {{- end }}
         volumeMounts:
+      {{- if .Values.customCA.enabled }}
```
danielezer: Is using the features for:
customVolumes and customVolumesMounts not good enough for you? You can achieve the same result with the more generic feature which already exists in the Artifactory chart. Not sure it makes sense to add this specifically.

stable/xray/templates/xray-server-statefulset.yaml
evannorstrand: This is as intended. If you check the README, I call it out specifically.

stable/artifactory/templates/artifactory-statefulset.yaml
evannorstrand: First: I am all for using existing functionality. It is more then good enough for me :) Before I put the time in for this PR, I contacted jFrog support, and they didn't mention the functionality and I couldn't find it in the documentation. 

Custom mounts in this case likely won't work based on the documentation available. Specifically because you need to add a single file to the folder, and the custom mount likely will over write all existing files. 

Adding a custom CA cert (single, not multiple) is required functionality for any environment that runs not connected to the internet.

My customers pay a ton of money for Artifactory, and this feature, whether added via this PR or another, is required for their continued use of Artifactory.

danielezer: @evannorstrand First of all. thank you for your time and your contribution. 
I apologise that the support team didn't tell you about this option, but it does exist and here is a link to the documentation: 
customVolumes: https://github.com/jfrog/charts/blob/master/stable/artifactory/README.md#custom-volumes

When you use `customVolumes` along with with `customVolumeMounts`, you can achieve your desired behaviour. 
Regarding your concern with mounting a single file - you can use the `subPath` option with the `customVolumeMounts` when mounting a configMap, same as you did.
Here's an example of a values file you can use:

```
artifactory:
  ## Set up your volume to the CA config map
  customVolumes: |
    - name: ca-configmap
      configMap:
        name: my-ca-configmap
  ## Mount the volume to the certs directory using the subPath method
  customVolumeMounts: |
    - name: ca-configmap
      mountPath: /etc/ssl/certs/custom-ca.crt
      subPath: custom-ca.crt
```

I agree that mounting a custom CA is an important feature, but if we already have a generic solution in place for that, I think it makes sense to use it.

You are correct though that is feature is missing from Xray. I think that this should definitely be added, if you're interested in suggesting a PR. If not, I'll be surely work on that next week :)

Please let me know if that solves your issue with Artifactory, if so, I will add that feature to Xray as well.

evannorstrand: For now I will close this PR, make the suggested changes in both this PR and the other PR that I submitted, and push a new PR. 

We need the functionality sooner rather then later for our customers to move forward, and i don't mind contributing back :)

PR #583, opened 2019-12-10
add toggles for http/https nginx service

### LINKED ISSUE #526: Artifactory in-accessible when spun up using AWS ELB Load Balancer

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): BUG REPORT

```
Client: &version.Version{SemVer:"v2.15.2", GitCommit:"8dce272473e5f2a7bf58ce79bb5c3691db54c96b", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.15.1", GitCommit:"cf1de4f8ba70eded310918a8af3a96bfe8e7683b", GitTreeState:"clean"}

Kubernetes EKS 1.14
```

**Which chart**: `artifactory-ha`

**What happened**: 
When using an AWS Load Balancer, classic or network, nginx starts but fails to serve content. 

By specifying [the supported annotations](https://kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#load-balancers), I can use an Amazon Load Balancer to front my nginx deployment and serve content via HTTP and/or HTTPS

Stand up an Artifactory deployment with the following nginx service annotations:

```
nginx:
  service:
    annotations:
      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: http
      service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:xx-xxxx-x:xxxxxxxxx:xxxxxxx/xxxxx-xxxx-xxxx-xxxx-xxxxxxxxx # replace this
```

* If you create the service using the `http` backend protocol, you'll be unable to access Artifactory via HTTPS because you'll be attempting to hit port 443 using the TCP protocol. 
* If you create the service using the `https` backend protocol, all of the instances will be marked as OutOfService because the health check will attempt to connect to port 80 (the port automatically selected for the health check) using the TLS protocol. 
* It would be great if we could select whether or not we want to serve up port 443.

andrewhertog:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

There is a kubernetes bug: https://github.com/kubernetes/kubernetes/issues/83070 which screws up the listeners on the ELB. We manually overroad the nginx service values to disable the http listener. This PR brings our manual change into compliance with your chart.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #526 

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.3.1] - Dec 10, 2019
+* Add support for toggleing nginx service ports
+
 ## [1.3.0] - Dec 1, 2019
 * Updated Artifactory version to 6.16.0
 
--- a/stable/artifactory-ha/Chart.yaml
+++ b/stable/artifactory-ha/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: artifactory-ha
 home: https://www.jfrog.com/artifactory/
-version: 1.3.0
+version: 1.3.1
 appVersion: 6.16.0
 description: Universal Repository Manager supporting all major packaging formats,
   build tools and CI servers.
--- a/stable/artifactory-ha/README.md
+++ b/stable/artifactory-ha/README.md
@@ -894,6 +894,12 @@ The following table lists the configurable parameters of the artifactory chart a
 | `nginx.https.internalPort` | Nginx service internal port           | `443`                           |
 | `nginx.replicator.internalPort` | Replicator service internal port | `6061`                          |
 | `nginx.replicator.externalPort` | Replicator service external port | `6061`                          |
+| `nginx.externalPortHttp` | DEPRECATED: Nginx service external port            | `80`                            |
+| `nginx.internalPortHttp` | DEPRECATED: Nginx service internal port            | `80`                            |
+| `nginx.externalPortHttps` | DEPRECATED: Nginx service external port           | `443`                           |
+| `nginx.internalPortHttps` | DEPRECATED: Nginx service internal port           | `443`                           |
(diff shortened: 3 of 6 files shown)


andrewhertog:
@eldada I think the postgres pod in the e2e tests is broken. I can't check the Travis CI build, but when I run the `make mac` locally, the postgres pod fails to start with the following fatal error: `FATAL:  could not open relation mapping file "global/pg_filenode.map": No such file or directory`

eldada:
> @eldada I think the postgres pod in the e2e tests is broken. I can't check the Travis CI build, but when I run the `make mac` locally, the postgres pod fails to start with the following fatal error: `FATAL: could not open relation mapping file "global/pg_filenode.map": No such file or directory`

The postgresql pod is part of the official postgresql chart we use. Nothing special on our side...

eldada:
> This breaks compatibility for the previous values. I don’t think we can do that. @eldada WDYT?

You are right @danielezer . 

@andrewhertog - can you leave the old config (with a deprecation note) along the new config?

andrewhertog:
> > This breaks compatibility for the previous values. I don’t think we can do that. @eldada WDYT?
> 
> You are right @danielezer .
> 
> @andrewhertog - can you leave the old config (with a deprecation note) along the new config?

Does bumping this to a minor release make more sense than a patch then?

eldada:
> > > This breaks compatibility for the previous values. I don’t think we can do that. @eldada WDYT?
> > 
> > 
> > You are right @danielezer .
> > @andrewhertog - can you leave the old config (with a deprecation note) along the new config?
> 
> Does bumping this to a minor release make more sense than a patch then?

I'd leave it as a patch.

andrewhertog:
Alright, i've added backwards compatibility. I've just added back the old values, and used an if/else to determine to use them or not. If both are there, it will use the new ones over the old ones.

Do i need to add a note in the README.md about deprecation too?

PR #584, opened 2019-12-10
Add support for rustic backups

evannorstrand:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

This PR allows the user to enable PV/PVC backups using Velero/Rustic. 

Tested on our integration systems, works fine, deployed to production.

This change should be merged AFTER my other PR, and will need some manual merging due to line number conflicts.

### DIFF
--- a/stable/artifactory/CHANGELOG.md
+++ b/stable/artifactory/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.3.2] - Dec 10, 2019
+* Added velero/rustic backup options to helm
+
 ## [8.3.0] - Dec 1, 2019
 * Updated Artifactory version to 6.16.0
 
--- a/stable/xray/templates/xray-server-statefulset.yaml
+++ b/stable/xray/templates/xray-server-statefulset.yaml
@@ -27,6 +27,9 @@ spec:
         component: {{ .Values.server.name }}
       annotations:
         checksum/config: {{ include (print .Template.BasePath "/xray-setup-conf.yaml") . | sha256sum }}
+      {{- if .Values.velero.restic.enabled}}
+        backup.velero.io/backup-volumes: "data-volume"
+      {{- end}}
     spec:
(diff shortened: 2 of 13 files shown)


evannorstrand:

stable/artifactory/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [8.3.2] - Dec 10, 2019
+* Added velero/rustic backup options to helm
```
danielezer: Is it rustic or restic?

stable/xray/templates/xray-server-statefulset.yaml
```diff
@@ -27,6 +27,9 @@ spec:
         component: {{ .Values.server.name }}
       annotations:
         checksum/config: {{ include (print .Template.BasePath "/xray-setup-conf.yaml") . | sha256sum }}
+      {{- if .Values.velero.restic.enabled}}
```
danielezer: I think this doesn’t make sense. We can’t add support with yaml flags to every annotation option k8s has to offer because it will never end. There are dozens of different annotations you can use to control stuff with k8s and we can’t support every single one with a dedicated flag. This is why we provide the generic option to provide any annotation you want. This is what you should use. I can see that such a feature is missing in the X-ray statefulset, so if you want to contribute a PR, this is what I would do. You can look at the ingress template we have as an example - we provide the option for the user to add any annotation they see fit

stable/artifactory/CHANGELOG.md
evannorstrand: Restic, sorry about that. I'll fix this.

stable/xray/templates/xray-server-statefulset.yaml
evannorstrand: Probably a better addition! I will look into adding a more generic update!

danielezer: Cool @evannorstrand, thanks for your contribution!

PR #588, opened 2019-12-12
Fix to the change adding the http(s) toggles in nginx

eldada:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Add missing change to nginx config as part of https://github.com/jfrog/charts/pull/583

### DIFF
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -48,7 +48,7 @@ ingress:
   # Used to create an Ingress record.
   hosts: []
   path: /
-  annotations:
+  annotations: {}
   # kubernetes.io/ingress.class: nginx
   # kubernetes.io/tls-acme: "true"
   labels:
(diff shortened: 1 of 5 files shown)


eldada:
@andrewhertog - FYI. We missed this.

stable/artifactory-ha/values.yaml
```diff
@@ -1007,8 +1007,12 @@ nginx:
...
     ## server configuration
     server {
-      listen {{ .Values.nginx.internalPortHttps }} ssl;
-      listen {{ .Values.nginx.internalPortHttp }} ;
+      {{- if .Values.nginx.https.enabled }}
+      listen {{ .Values.nginx.https.internalPort }} ssl;
+      {{- end }}
+      {{- if .Values.nginx.http.enabled }}
+      listen {{ .Values.nginx.http.internalPort }} ;
+      {{- end }}
```
andrewhertog: Thanks for catching this.

If i'm not mistaken though, this isn't backwards compatible to the old 1.3.0 values?

eldada: You are **right**! I will fix this too. Thx for noticing!

PR #606, opened 2020-01-04
artifactory-ha: Support for SSL termination in Nginx service layer. 

Satheesh-Balachandran:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

We currently don't have the support to terminate/offload SSL in the LoadBalancer layer of Nginx.

### DIFF
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -1066,18 +1066,6 @@ nginx:
       if ($http_x_forwarded_proto = '') {
         set $http_x_forwarded_proto  $scheme;
       }
-      {{- if .Values.nginx.service.ssloffload }}
-      if ($host = $server_addr) {
-        rewrite ^/artifactory/webapp/(.*) http://$host:8081/artifactory/webapp/$1 redirect;
-      }
-      if ($http_x_forwarded_proto != "https") {
-        rewrite ^/artifactory/webapp/(.*) https://$host/artifactory/webapp/$1 redirect;
-        rewrite ^/$ $scheme://$host/artifactory/webapp/#/home redirect;
-      }
-      if ($http_x_forwarded_proto = "https") {
-        rewrite ^/$ https://$host/artifactory/webapp/#/home redirect;
-      }
-      {{- end }}
       ## Application specific logs
       ## access_log /var/log/nginx/artifactory-access.log timing;
       ## error_log /var/log/nginx/artifactory-error.log;
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -192,12 +192,30 @@ artifactory:
   # - artifactory.log
   # - build_info_migration.log
 
+  # Loggers containers resources
+  loggersResources: {}
+  #  requests:
+  #    memory: "64Mi"
+  #    cpu: "25m"
+  #  limits:
+  #    memory: "128Mi"
+  #    cpu: "50m"
+
   # Sidecar containers for tailing Tomcat (catalina) logs
   catalinaLoggers: []
   # - catalina.log
   # - host-manager.log
   # - localhost.log
   # - manager.log
+
+  # Tomcat (catalina) loggers resources
+  catalinaLoggersResources: {}
+  #  requests:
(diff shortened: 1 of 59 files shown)


stable/artifactory-ha/values.yaml
```diff
@@ -1064,6 +1066,18 @@ nginx:
       if ($http_x_forwarded_proto = '') {
         set $http_x_forwarded_proto  $scheme;
       }
+      {{- if .Values.nginx.service.ssloffload }}
+      if ($host = $server_addr) {
+        rewrite ^/artifactory/webapp/(.*) http://$host:8081/artifactory/webapp/$1 redirect;
```
danielezer: Change 8081 to `{{ .Values.artifactory.externalPort }}`

stable/artifactory-ha/values.yaml
```diff
@@ -1064,6 +1066,18 @@ nginx:
       if ($http_x_forwarded_proto = '') {
         set $http_x_forwarded_proto  $scheme;
       }
+      {{- if .Values.nginx.service.ssloffload }}
+      if ($host = $server_addr) {
```
danielezer: In what case would `$host` be different from the `$server_addr` ? From looking at the nginx docs, seems like they are the same. But maybe I'm missing something..?

stable/artifactory-ha/values.yaml
```diff
@@ -1064,6 +1066,18 @@ nginx:
...
+        rewrite ^/artifactory/webapp/(.*) http://$host:8081/artifactory/webapp/$1 redirect;
+      }
+      if ($http_x_forwarded_proto != "https") {
+        rewrite ^/artifactory/webapp/(.*) https://$host/artifactory/webapp/$1 redirect;
+        rewrite ^/$ $scheme://$host/artifactory/webapp/#/home redirect;
+      }
+      if ($http_x_forwarded_proto = "https") {
+        rewrite ^/$ https://$host/artifactory/webapp/#/home redirect;
```
danielezer: Doesn't this mean that every request that comes in with https will be redirected to the home page?

stable/artifactory-ha/values.yaml
```diff
@@ -1085,13 +1099,25 @@ nginx:
...
+        {{- else }}
         proxy_set_header    X-Artifactory-Override-Base-Url $http_x_forwarded_proto://$host:$server_port/artifactory;
         proxy_set_header    X-Forwarded-Port  $server_port;
+        {{- end }}
         proxy_set_header    X-Forwarded-Proto $http_x_forwarded_proto;
         proxy_set_header    Host              $http_host;
         proxy_set_header    X-Forwarded-For   $proxy_add_x_forwarded_for;
       }
     }
+    {{- if .Values.nginx.service.ssloffload }}
```
danielezer: Why is this needed?

Satheesh-Balachandran: @eldada @danielezer If SSL is terminated in AWS load balancer with ACM certificate, only the HTTP request goes to the service.

stable/artifactory-ha/values.yaml
Satheesh-Balachandran: $server_addr is the address of the localhost in which Nginx runs. This is to make sure the Nginx readinessProbe/livenessProbe doesn't break. Nginx checks for artifactory health in https protocol. Another server block to listen in 8081 was configured which redirects to the application.

Satheesh-Balachandran: @danielezer Suggest a better way to handle the Nginx readiness/liveness

stable/artifactory-ha/values.yaml
Satheesh-Balachandran: @danielezer Removed the Force SSL redirect part.

eldada:
@Satheesh-Balachandran - can you wait for the new unified charts to apply this change there?

Satheesh-Balachandran:
@eldada  Sure thanks!

chukka:
@Satheesh-Balachandran  I'm closing this ,Feel free to raise a new PR for these changes

PR #615, opened 2020-01-15
Refined Pod Disruption Budgets (Separated for Nginx and Artifactory pods)

### LINKED ISSUE #613: Pod Disruption Budget uses (potentially) incorrect labels
Yes

**This is a FEATURE REQUEST** 
This is potentially a feature request, although I am unsure why in its current state it would be useful

**Describe *in detail* the feature/behavior/change you'd like to see.**

For the following chart artifactory-node-pdb.yaml the labels the Pod Disruption Budget match on are shared with the Nginx pods. 
This means that you can't just simply define minAvailable as 1 to ensure that either the Primary or Member node remains active. 
I would like the ability to just match on the label "component" as the Primary and Member nodes have "artifactory-ha" as their component whereas the Nginx pods have "nginx" 

**Which chart**:
artifactory-node-pdb.yaml

I may not be understanding something - I am unsure how to make a PR for this issue as I am not sure how to pass in keys rather than values (as in, have the user of the chart decide which labels they would like to base the PDB off when deploying the chart)

DNewt:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated
- [-] Variables and other changes are documented in the README.md
Unsure if this should be in the ReadMe or not, I can't see anything about pod disruption budgets in there at the moment

This PR was required for being able to specifically target just the Member and Primary Artifactory pods with a pod disruption budget - otherwise the Nginx pods are included. 

This PR adds another template for specifically targeting the Nginx pods.

Both of these charts target the specific pods by using the component label (which is nginx for Nginx and artifactory-ha for the Primary and Member nodes)

Fixes #613 

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [1.4.2] - Jan 22, 2020
+* Refined pod disruption budgets to separate nginx and Artifactory pods
+
 ## [1.4.1] - Jan 19, 2020
 * Fix replicator port config in nginx replicator configmap
 
--- a/stable/artifactory-ha/Chart.yaml
+++ b/stable/artifactory-ha/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: artifactory-ha
 home: https://www.jfrog.com/artifactory/
-version: 1.4.1
+version: 1.4.2
 appVersion: 6.17.0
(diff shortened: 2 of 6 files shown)


DNewt:
Changes made :)

DNewt:
@danielezer / @eldada 
Hey for some reason this change has been removed in https://github.com/jfrog/charts/pull/666
(@chukka)
Unsure why?

chukka:
@DNewt Can you please share more info like chart version you are using (recommended to raise as an issue if it doesn't work )

FYI, https://github.com/jfrog/charts/blob/master/stable/artifactory-ha/values.yaml#L1134-L1147

PR #628, opened 2020-01-28
Fix HA Node ID

AlexanderRMcNeill:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

This PR prevents the HA node ID exceeding 40 characters when a release name is over 16 characters. Exceeding a 40 character HA node ID results in errors from Nuget acquiring an HA lock. 

This PR fixes this by using the pod UID for the HA node ID which is always 36 characters instead of using the pod name which is variable based on the release name.

**Which issue this PR fixes** *
This doesn't fix the following issue but avoids it:
https://www.jfrog.com/jira/browse/RTFACT-18851

### DIFF
--- a/stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
+++ b/stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
@@ -268,7 +268,9 @@ spec:
         - name: HA_MEMBERSHIP_PORT
           value: "{{ .Values.artifactory.membershipPort }}"
         - name: HA_NODE_ID
-          value: "{{ trimPrefix .Release.Name metadata.name }}"
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.uid
         - name: ART_PRIMARY_BASE_URL
           value: 'http://{{ template "artifactory-ha.primary.name" . }}:{{ .Values.artifactory.internalPort }}/artifactory'
       {{- if eq .Values.artifactory.persistence.type "file-system" }}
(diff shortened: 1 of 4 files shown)


stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -270,7 +270,7 @@ spec:
         - name: HA_NODE_ID
           valueFrom:
             fieldRef:
-              fieldPath: metadata.name
+              fieldPath: metadata.uid
```
AlexanderRMcNeill: I'm not sure if there is an issue with the HA Node ID changing every time a new pod comes up. 

I initially tried to trim the release name off the pod name but had no luck due to the pod name not being available when the helm renders the templates.

eldada: It is a problem as this id is used to register the node in the cluster and each node consumes a single license. After each pod redeployment, you'll get a new node registered in the cluster, which will show up on the list of instances and will not get a license. 
Need to look at maybe using a template for this.

AlexanderRMcNeill: Looks like sprig has a uuid function (https://masterminds.github.io/sprig/uuid.html) that might also work. This means you would get a new HA Node ID every time you release which is slightly better.

eldada: Still - means every upgrade will force new IDs. We need to persist the IDs. One of the reasons for using a statefulset.

AlexanderRMcNeill: Ok, I'll create an issue for this so it is documented for anyone who runs into it while we are trying to find as solution 😃 

This is not a big deal for us now that we know what was causing our nuget issues but had us scratching our heads for awhile.

PR #672, opened 2020-03-13
artifactory: Fixed: typo logabck.xml -> logback.xml

joe-sonrichard:
Context suggests it is always supposed to be `logback.xml` and that there is not another file, called `logabck.xml`, which is being referenced.

#### PR Checklist
Fixes a typo. Though minor, it could be confusing to someone if they see two different spellings and assume there are two files in play.

### DIFF
--- a/stable/artifactory/CHANGELOG.md
+++ b/stable/artifactory/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [9.0.25] - Mar 16, 2020
+* Fix docs typo logabck.xml -> logback.xml
+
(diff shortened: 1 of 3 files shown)


rimusz:
Thanks for the PR.
Please bump chart version and update changelog file as well.

joe-sonrichard:
Got it, I've bumped the changlog and version. First time I did that I found out version 9.0.25 had been taken so I had to change it again after re-basing off master. Used a force push to try to clean it up a bit.

eldada:
@joe-sonrichard - can you rebase again and resolve conflict (another merge was just made). Thx in advance and sorry for the hassle.

eldada:
I will fix this with #681 to save you the trouble.

PR #673, opened 2020-03-13
artifactory: Changed '-hex 16' -> '-hex 32' to agree with rest of doc

joe-sonrichard:
#### PR Checklist
Minor consistency fix. 
 - Table says keys are "**32** hex characters"
 - Instructions when creating [master key](https://github.com/jfrog/charts/blob/master/stable/artifactory/README.md#create-a-unique-master-key) and [join-key](https://github.com/jfrog/charts/blob/master/stable/artifactory/README.md#create-a-unique-join-key) say to create one with **32** hex characters.
 - Example command in table creates a key with **16** characters.

**Change:** command in table now creates a key with **32** characters.

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [2.0.30] - Mar 16, 2020
+* Add Unsupported message from 6.18 to 7.2.x (migration)
+
 ## [2.0.29] - Mar 11, 2020
 * Upgrade Docs update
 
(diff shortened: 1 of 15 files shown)


rimusz:
Thanks for the PR.
Please bump chart version and update changelog file as well.

JfrogPrasanna:
@joe-sonrichard Thanks for the PR. The issue is that we need to align this  with the 6.x where the support for 16b so upgrade can  be seamless. We will  look at migration script to handle this and take your  PR as well. Thank you.

joe-sonrichard:
Got it, I've bumped the changelog and version. I made the version on higher than I did in PR #672 since it sounds like that would be merged in sooner.

Also rebased off master.

chukka:
Closing this and raised a PR https://github.com/jfrog/charts/pull/744

PR #715, opened 2020-03-24
Pipelines chart v1.0.33

rimusz:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

### DIFF
--- a/stable/pipelines/templates/pipelines-hpa.yaml
+++ b/stable/pipelines/templates/pipelines-hpa.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.pipelines.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ include "pipelines.services.name" . }}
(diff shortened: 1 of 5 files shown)


stable/pipelines/values.yaml
```diff
@@ -617,6 +625,9 @@ pipelines:
       defaultMinionInstanceSize: "c4.large"
       allowDynamicNodes: true
       allowCustomNodes: true
+      {{- range $key, $value := .Values.runtimeOverride }}
```
danielezer: Wouldn't it be easier here to just use `toYaml`?

rimusz: I copied what @khazina had done

stable/pipelines/templates/pipelines-hpa.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.pipelines.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
```
eldada: Can we use `v1`?

stable/pipelines/values.yaml
eldada: Should probably be `with` and `tpl`. We can change it later on...

stable/pipelines/templates/pipelines-hpa.yaml
rimusz: I set the same version we have in Xray

PR #719, opened 2020-03-25
Fix volume name used by filebeat container

### LINKED ISSUE #718: wrong volume for filebeat
this is a bug report.

wrong volume for the statefulset has been set instead of `volume` it should be `artifactory-volume` on line 452 https://github.com/jfrog/charts/blob/master/stable/artifactory/templates/artifactory-statefulset.yaml

eldada:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Fix volume name used by filebeat container. Was wrongly using `volume` instead of `artifactory-volume`.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #718 

### DIFF
--- a/stable/artifactory/CHANGELOG.md
+++ b/stable/artifactory/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [9.1.4] - Mar 25, 2020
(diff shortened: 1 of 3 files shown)


eldada:
> What about the reference to that volume?

This is the reference.

PR #731, opened 2020-03-30
[artifactory-ha] Add init resources + version customisation for migration pod

KierranM:
#### PR Checklist
- [X] Chart Version bumped
- [X] CHANGELOG.md updated
- [X] Variables and other changes are documented in the README.md

Three init containers have been added that currently don't respect the `initContainers.resources` value. In an environment where there are default resources this causes the migration pod to be unable to start Artifactory (for example, our default requests/limits are 64Mi mem and 10m CPU).

It's also possible to override the artifactory version that is used in _almost_ every section of the chart, except when the migration script is called. This has been updated to use the same defaulting mechanism used everywhere else

EDIT: I also found that you can't specify an external license secret. The volume was defined twice if you did, so I've fixed that


eldada:
Please rebase and resolve conflicts.

KierranM:
Rebased 😄

PR #742, opened 2020-04-01
[artifactory-ha] External joinKey secret + node statefulset resources

KierranM:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

This makes the template render the script to copy the join key when it is provided by an external secret. Currently this whole block is ignored if `artifactory.joinKey` is not set to any value, so the value in the secret is never applied.

It also applies a couple of the changes from #731 to the node statefulset (I didn't realise the node had it at the time)

### DIFF
--- a/stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
+++ b/stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
@@ -185,7 +185,7 @@ spec:
           mkdir -p {{ .Values.artifactory.persistence.mountPath }}/log;
           bash $scriptsPath/migrationStatus.sh {{ default .Chart.AppVersion .Values.artifactory.image.version }} > >(tee {{ .Values.artifactory.persistence.mountPath }}/log/helm-migration.log) 2>&1;
         resources:
-{{ toYaml .Values.initContainers.resources | indent 10 }}
+{{ toYaml .Values.artifactory.node.resources | indent 10 }}
         env:
         {{- if .Values.database.secrets.user }}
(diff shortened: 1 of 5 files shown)


stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -183,7 +183,9 @@ spec:
           cp -fv /tmp/migrationHelmInfo.yaml $scriptsPath/migrationHelmInfo.yaml;
           cp -fv /tmp/migrationStatus.sh $scriptsPath/migrationStatus.sh;
           mkdir -p {{ .Values.artifactory.persistence.mountPath }}/log;
-          bash $scriptsPath/migrationStatus.sh {{ .Chart.AppVersion }} > >(tee {{ .Values.artifactory.persistence.mountPath }}/log/helm-migration.log) 2>&1;
+          bash $scriptsPath/migrationStatus.sh {{ default .Chart.AppVersion .Values.artifactory.image.version }} > >(tee {{ .Values.artifactory.persistence.mountPath }}/log/helm-migration.log) 2>&1;
+        resources:
+{{ toYaml .Values.initContainers.resources | indent 10 }}
```
eldada, line 188: The migration container will not work with `initContainers.resources` as it actually starts Artifactory. Need to provide it with the same resources as the node is set: `artifactory.node.resources`.

KierranM: Sweet as. I'll update the primary migration container as well

PR #774, opened 2020-04-06
Fix cache size

camertron:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

I believe the cache size currently set in values.yaml is 50gb (assuming units of bytes), which is larger than the persistent volume claim size of 20gb. There's an extra zero. This caused our Artifactory instance to effectively "crash" today since it couldn't write anything else to the cache 😢


chukka:
Hi @camertron Thanks for the PR. Looks like you using old charts (can you please rebase and update)

Can I bother you to do 2 things?

1. Change the destination branch to be inbound-from-forks. We are now using a two stage contribution flow. Once merged, we'll take into master. (https://github.com/jfrog/charts#contributing-to-jfrog-charts)
2.Add the same change to the artifactory-ha chart.

Also, please confirm which chart version of artifactory you are using 


**Note:** we have 2 chart versions (`master` branch has unified (latest) charts and `pre-unified-platform` branch has old charts)

camertron:
@chukka ahh ok that explains the merge conflicts. I was really surprised to see them after branching off a 5-minute old master!

PR #826, opened 2020-04-16
Custom volume mount for initcontainer

Satheesh-Balachandran:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Custom volume mounts in migration init container to make use of files like postgres SSL certificate which is required on startup


eldada:
> What about the customVolumes? Not sure it makes sense to only have the customVolumeMounts

These are using the volumes defined on the whole statefulset spec level. The volume mounts are per container.

danielezer:
Oh ok than

PR #830, opened 2020-04-20
Fix broken support if `redis.password` not explicitly set

chukka:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Fix broken support if `redis.password` not explicitly set


stable/distribution/templates/distribution-statefulset.yaml
```diff
@@ -186,6 +193,13 @@ spec:
           {{- end }}
             /opt/jfrog/distribution/app/bin/wrapper.sh;
         env:
+        - name: JF_SHARED_REDIS_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "distribution.fullname" . }}-redis-secret
```
rimusz, line 199: please add an external secret support for Redis password

stable/distribution/values.yaml
```diff
@@ -132,6 +132,7 @@ redis:
     pullPolicy: IfNotPresent
   port: 6379
   password: ""
+  # existingSecret: <name-of-existing-secret>
```
rimusz, line 135: add in comment an example of the key:
`key: redis-password`

stable/distribution/templates/redis-connection-secret.yaml
```diff
@@ -1,17 +0,0 @@
-apiVersion: v1
```
danielezer, line 1: Why has this been deleted? Where would the password be stored in a case where the user didn't provide an existing secret?

stable/distribution/README.md
```diff
@@ -259,6 +259,7 @@ The following table lists the configurable parameters of the distribution chart
 | `postgresql.affinity`                           | PostgreSQL node affinity                                               | `{}`                                                               |
 | `postgresql.tolerations`                        | PostgreSQL node tolerations                                            | `[]`                                                               |
 | `redis.password`                                | Redis password                                                         | ` `                                                                |
+| `redis.existingSecret`                          | Redis existingSecret for password                                      | ` `                                                                |
```
danielezer, line 262: I think we should keep the convention we have in other charts, which is `secretName`

stable/distribution/templates/redis-connection-secret.yaml
chukka: I moved this logic to `distribution-secrets.yaml` file

stable/distribution/README.md
chukka: @danielezer I took the convention as per upstream chart 
https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml#L205

danielezer: ok, makes sense

PR #883, opened 2020-05-06
[ mission-control ] - Add support to external database secrets

chukka:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

-  Add support to external database secrets
-  Add support to use single user/password for all services


stable/mission-control/templates/mission-control-statefulset.yaml
```diff
@@ -148,13 +148,8 @@ spec:
         - name: PGPASSWORD
           valueFrom:
             secretKeyRef:
-              {{- if .Values.postgresql.postgresqlPassword }}
-              name: {{ template "mission-control.fullname" . }}-db-cred
-              key: postgres-password
-            {{- else }}
               name: {{ .Release.Name }}-postgresql
```
danielezer, line 151: This won’t work with an external posters server.

stable/mission-control/templates/postgresql-secrets.yaml
```diff
@@ -0,0 +1,33 @@
...
+  {{- end }}
+  {{- if .Values.postgresql.db.jfscPassword }}
+  jfscPassword: {{ .Values.postgresql.db.jfscPassword | b64enc | quote }}
+  {{- else }}
+  jfscPassword: {{ randAlphaNum 10 | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.postgresql.db.jfexPassword }}
+  jfexPassword: {{ .Values.postgresql.db.jfexPassword | b64enc | quote }}
+  {{- else }}
+  jfexPassword: {{ randAlphaNum 10 | b64enc | quote }}
```
danielezer, line 26: Why are all of these here if you removed it from the values?

stable/mission-control/templates/mission-control-statefulset.yaml
chukka: This condition is only for {{- if .Values.postgresql.enabled }}

stable/mission-control/templates/postgresql-secrets.yaml
chukka: These are for internal postgres ~ postgresql.enabled=true, removed only for external database password

danielezer:
It doesn’t make sense in my opinion. It should either be one secret in both cases or all secrets in both cases...

chukka:
@danielezer Updated changes for both internal/external databases

PR #931, opened 2020-06-05
[artifactory-ha]Service annotations for members and primary.

astrowicked:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

When using the loadbalancer type for artifactory.services, the non-primary members service is the only service that has annotations applied to it. We utilize the annotations to specify an internal AWS LB. This allows us to use our inter-datacenter load balancers to route traffic over DirectConnect to our EKS Artifactory as well as our On-Prem Artifactory instances.

We utilize Istio as public ingress which is why the nginx and kubernetes ingress is disabled(sidecar injection does not seem to play nice with nginx). 

I ran make lint successfully prior to opening the PR and deployed to a test EKS cluster with the "service.beta.kubernetes.io/aws-load-balancer-internal=true" annotation set on artifactory.services.annotations and confirmed both primary and member services were provisioned with and internal AWS ELB as expected.

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [2.5.2] - Jun 5, 2020
+* Applies service annotations to primary service (Previously only applied to members service)
+
 ## [2.5.1] - Jun 5, 2020
 * Fixes broken PDB issue upgrading from 6.x to 7.x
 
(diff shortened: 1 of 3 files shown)


astrowicked:
@amithins @danielezer @eldada @chukka @rimusz

astrowicked:
I guess I misunderstood the use for the primary service. If it's intention is meant to be internal only for readiness check, wouldn't it be worth not allowing the service type to be set to LoadBalancer, or allowing it to be set as such but with annotations to have the LB on a private subnet?

When I set the service type as LoadBalancer on an EKS deployment, I would get an internal subnet LB for the members service but then a publicly exposed LB for the primary service because it did not honor the annotations. 

Is that the intention? Sorry if I'm missing something.

danielezer:
@astrowicked This is actually a bug... The primary service should not be used by users, it is only used for internal purposes. It would be great if you could change the primary service to always be:
```
spec:
  type: ClusterIP
```
Instead of using the parameter from the values file. 
Thanks for your contribution!

astrowicked:
Reopening under fresh fork and proper issue description

PR #973, opened 2020-06-26
[Xray] Router container prestartcommand implementation/fix

peters95:
#### PR Checklist
- [X] Chart Version bumped
- [X] CHANGELOG.md updated
- [X] Title of the PR starts with chart name (e.g. `[artifactory]`)

Currently only 4 out of 5 xray container respect the prestartcommand that is submitted via the values.yaml this will address this issue and now the same mechanism will be used in all 5 out of 5 containers that xray starts in the pod.

Fixed our GKE marketplace issues in partner engineering such as PTRENG-270 Xray IaaS (epic)

This is needed to resolve problems in our GKE marketplace combined artifactory + xray listing as we need xray to "wait" for artifactory to start. By fixing this in all 5 containers we are then able to have this combined product listing work in the GKE marketplace.

### DIFF
--- a/stable/xray/templates/xray-statefulset.yaml
+++ b/stable/xray/templates/xray-statefulset.yaml
@@ -120,6 +120,15 @@ spec:
       - name: {{ .Values.router.name }}
         image: '{{ .Values.router.image.repository }}:{{ default .Chart.AppVersion .Values.router.image.version }}'
         imagePullPolicy: {{ .Values.router.image.imagePullPolicy }}
+        command:
+          - '/bin/sh'
+          - '-c'
+          - >
+          {{- with .Values.common.preStartCommand }}
+            echo "Running custom common preStartCommand command";
+            {{ tpl . $ }};
+          {{- end }}
+            /opt/jfrog/router/app/bin/entrypoint-router.sh;
         ports:
           - name: router
             containerPort: {{ .Values.router.internalPort }}
--- a/stable/xray/CHANGELOG.md
+++ b/stable/xray/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Xray Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [3.5.1] - Jun 25, 2020
+* Added prestartcommand to router container to match same mechanism in all other xray containers 
+
 ## [3.5.0] - Jun 22, 2020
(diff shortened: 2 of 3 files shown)


stable/xray/templates/xray-statefulset.yaml
```diff
@@ -120,6 +120,15 @@ spec:
...
         imagePullPolicy: {{ .Values.router.image.imagePullPolicy }}
+        command:
+          - '/bin/sh'
+          - '-c'
+          - >
+          {{- with .Values.common.preStartCommand }}
+            echo "Running custom common preStartCommand command";
+            {{ tpl . $ }};
+          {{- end }}
+            /opt/jfrog/router/app/bin/entrypoint-router.sh;
```
eldada, line 131: Use `exec` before. Like we have in artifactory charts.

stable/xray/templates/xray-statefulset.yaml
```diff
@@ -120,6 +120,15 @@ spec:
       - name: {{ .Values.router.name }}
         image: '{{ .Values.router.image.repository }}:{{ default .Chart.AppVersion .Values.router.image.version }}'
         imagePullPolicy: {{ .Values.router.image.imagePullPolicy }}
+        command:
+          - '/bin/sh'
```
eldada, line 124: Change `sh` to `bash` (make sure we have it in the image first).

peters95: Hi Eldada I copied the same configuration that all the other xray prestart command use. Can you explain why you want just this one to be different than the other 4? This is not the artifactory chart.

stable/xray/templates/xray-statefulset.yaml
peters95: Please see comment below. I would expect this to be the same as all the other prestartcommands but we could apply this throughout if necessary. The goal of this PR was only to make router match the other 4 containers. It seems like somebody forgot about this container to be honest.

stable/xray/templates/xray-statefulset.yaml
eldada: It's needed to properly pass SIGTERM to entrypoint. Indeed - this should be for all containers. I will send you more details in a DM. Leave this for now.

PR #982, opened 2020-06-29
[artifactory] Add replication ingress object

anastasiagrinman:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated


stable/artifactory-ha/README.md
```diff
@@ -1475,6 +1476,32 @@ and running:
 helm upgrade --install artifactory-ha --namespace artifactory-ha jfrog/artifactory-ha -f artifactory-ha-values.yaml
 ```
 
+### Dedicated Ingress object for replicator service
+
+You have the option to add additional ingress object to the Replicator service. An example for this use case can be routing the /replicator/ path to Artifactory.
```
danielezer, line 1481: @anastasiag the use here can also be covered by using ingress.additionalRules which is currently supported. Can you explain why we need a separate ingress resource for this?

anastasiagrinman: In case you want to override specific annotations per location, you cannot do it on the same configuration-snippet, since it covers all the PATH'es locations from the same ingress object.
I.E. that's why we wanted to have an option to add a new Ingress object that have specific overrides annotations.

danielezer: And the chunked encoding thing cant be there for Artifactory as well? Does it have to be for replicator only?

PR #1007, opened 2020-07-09
Make yaml and markdown files non-executable.

kristinnardal2:
There is no reason for them to be executable. I ran the following
command inside the stable directory

```bash
$ find . -executable -type f | xargs chmod 0644
```

No need to bump any chart versions or anything, this is just file permissions.


chukka:
@kristinnardal2  Thanks for Info. we will be pushing these changes in https://github.com/jfrog/charts/pull/1023 and https://github.com/jfrog/charts/pull/1020

I'm closing this PR

PR #1066, opened 2020-08-12
[Pipelines] Add support for external rabbitmq and redis

vivekkumar-git:
#### PR Checklist
- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated
- [ ] Title of the PR starts with chart name (e.g. `[artifactory]`)

### DIFF
--- a/stable/pipelines/values.yaml
+++ b/stable/pipelines/values.yaml
@@ -858,7 +858,7 @@ global:
     ssl: false
     ## PostgreSQL password using existing secret
     # existingSecret: secret
-  
+
   rabbitmq:
     internal_ip: "{{ .Release.Name }}-rabbitmq"
     msg_hostname: "{{ .Release.Name }}-rabbitmq"
--- a/stable/pipelines/values.yaml
+++ b/stable/pipelines/values.yaml
@@ -504,7 +504,7 @@ pipelines:
         defaultExchange: {{ .Values.rabbitmq.root_vhost_exchange_name }}
         amqpVhost: {{ .Values.rabbitmq.build_vhost_name}}
         amqpRootVhost: {{ .Values.rabbitmq.root_vhost_name }}
-        protocol: {{ .Values.rabbitmq.protocol | default "amqps" }}
+        protocol: {{ .Values.rabbitmq.protocol }}
         {{- end }}
         queues:
           - "core.pipelineSync"
@@ -535,7 +535,7 @@ pipelines:
           {{- else if .Values.rabbitmq.enabled }}
           url: amqp://{{ tpl .Release.Name . }}-rabbitmq
           {{- else }}
-          url: {{ .Values.rabbitmq.protocol | default "amqps" }}://{{ tpl .Values.rabbitmq.msg_hostname . }}:{{ .Values.rabbitmq.port }}
+          url: {{ .Values.rabbitmq.protocol }}://{{ tpl .Values.rabbitmq.msg_hostname . }}:{{ .Values.rabbitmq.port }}
           {{- end }}
           rootUrl: ""
           adminUrl: ""
@@ -902,6 +902,7 @@ postgresql:
 ##
 rabbitmq:
   enabled: true
+  protocol: amqps
   replicas: 1
 
   rabbitmq:
--- a/stable/pipelines/values.yaml
+++ b/stable/pipelines/values.yaml
@@ -493,7 +493,7 @@ pipelines:
         amqpVhost: pipelines
         amqpRootVhost: pipelinesRoot
         {{- else }}
-        ip: {{ .Values.rabbitmq.internal_ip }}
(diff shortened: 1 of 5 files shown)


stable/pipelines/values.yaml
```diff
@@ -502,7 +502,7 @@ pipelines:
         defaultExchange: {{ .Values.rabbitmq.root_vhost_exchange_name }}
         amqpVhost: {{ .Values.rabbitmq.build_vhost_name}}
         amqpRootVhost: {{ .Values.rabbitmq.root_vhost_name }}
-        protocol: amqps
+        protocol: amqp
```
rimusz, line 505: Please an option to set `amqp` or `amqps`

stable/pipelines/values.yaml
```diff
@@ -533,7 +533,7 @@ pipelines:
           {{- else if .Values.rabbitmq.enabled }}
           url: amqp://{{ tpl .Release.Name . }}-rabbitmq
           {{- else }}
-          url: amqps://{{ .Values.rabbitmq.msg_hostname }}:{{ .Values.rabbitmq.port }}
+          url: amqp://{{ tpl .Values.rabbitmq.msg_hostname . }}:{{ .Values.rabbitmq.port }}
```
rimusz, line 536: same here, please an option to set `amqp` or `amqps`

stable/pipelines/values.yaml
jatindogra: I agree this should not be a default, there should be an option to set `amqp` or `amqps`

eldada: Please make it as similar to Xray as possible. We would like to eventually share config.
Also, it might be a good idea to add support for the `global` option here... See https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values

rimusz:
The changes should be documented

rimusz (changes requested):
The changes should be documented

stable/pipelines/values.yaml
```diff
@@ -504,7 +504,7 @@ pipelines:
         defaultExchange: {{ .Values.rabbitmq.root_vhost_exchange_name }}
         amqpVhost: {{ .Values.rabbitmq.build_vhost_name}}
         amqpRootVhost: {{ .Values.rabbitmq.root_vhost_name }}
-        protocol: amqps
+        protocol: {{ .Values.rabbitmq.protocol | default "amqps" }}
```
danielezer, line 507: Add this to the values instead of using | default.
This causes redundant duplications

stable/pipelines/values.yaml
```diff
@@ -535,7 +535,7 @@ pipelines:
           {{- else if .Values.rabbitmq.enabled }}
           url: amqp://{{ tpl .Release.Name . }}-rabbitmq
           {{- else }}
-          url: amqps://{{ .Values.rabbitmq.msg_hostname }}:{{ .Values.rabbitmq.port }}
+          url: {{ .Values.rabbitmq.protocol | default "amqps" }}://{{ tpl .Values.rabbitmq.msg_hostname . }}:{{ .Values.rabbitmq.port }}
```
danielezer, line 538: Same here

stable/pipelines/values.yaml
```diff
@@ -567,8 +567,6 @@ pipelines:
       redis:
         ip: {{ .Release.Name }}-redis-master
         port: 6379
-        image: "jfrog/pipelines-redis"
```
danielezer, line 570: Why was this removed?

danielezer (changes requested):

stable/pipelines/values.yaml
chukka: This is not needed .. redundant

stable/pipelines/values.yaml
rahulsadanandan: @danielezer , This is when we are using an external rabbitmq. If we removed the default (amqps) which was there before. would that break any existing deployments , if they dont provide the protocol. ?

danielezer: @rahulsadanandan I'm not asking you to remove the protocol. I'm asking that you add `rabbitmq.protocol=amqps` to the values.yaml and that will remove the need to use `| default "amqps"`

rahulsadanandan: Changed.

stable/pipelines/values.yaml
rahulsadanandan: changed

danielezer (approved):

rimusz:
Also please move external rabbitmq/redis settings to `global:` to be consistent, as there external `postgres` gets set

jatindogra (approved):

rahulsadanandan (approved):

chukka (approved):

chukka:
> The changes should be documented

@rimusz  Done

PR #1094, opened 2020-09-03
[artifactory] [artifactory-ha] feat: Expose Prometheus metrics

rpf3:
This commit adds the configuration required to export Prometheus metrics
by creating a new service and service monitor as well as installing
Fluentd as a sidecar to the Artifactory container.

#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Adds support for exposing Prometheus metrics according to JFrog documentation using Fluentd.


stable/artifactory-ha/README.md
```diff
@@ -756,7 +756,7 @@ database:
...
 artifactory:
-  preStartCommand: "mkdir -p /opt/jfrog/artifactory/var/bootstrap/artifactory/tomcat/lib; cd /opt/jfrog/artifactory/var/bootstrap/artifactory/tomcat/lib && wget -O instantclient-basic-linux.x64-19.6.0.0.0dbru.zip https://download.oracle.com/otn_software/linux/instantclient/19600/instantclient-basic-linux.x64-19.6.0.0.0dbru.zip && unzip -jn instantclient-basic-linux.x64-19.6.0.0.0dbru.zip && wget -O libaio1_0.3.110-3_amd64.deb http://ftp.br.debian.org/debian/pool/main/liba/libaio/libaio1_0.3.110-3_amd64.deb &&  dpkg-deb -x libaio1_0.3.110-3_amd64.deb . && cp lib/x86_64-linux-gnu/* ."  
+  preStartCommand: "mkdir -p /opt/jfrog/artifactory/var/bootstrap/artifactory/tomcat/lib; cd /opt/jfrog/artifactory/var/bootstrap/artifactory/tomcat/lib && wget -O instantclient-basic-linux.x64-19.6.0.0.0dbru.zip https://download.oracle.com/otn_software/linux/instantclient/19600/instantclient-basic-linux.x64-19.6.0.0.0dbru.zip && unzip -jn instantclient-basic-linux.x64-19.6.0.0.0dbru.zip && wget -O libaio1_0.3.110-3_amd64.deb http://ftp.br.debian.org/debian/pool/main/liba/libaio/libaio1_0.3.110-3_amd64.deb &&  dpkg-deb -x libaio1_0.3.110-3_amd64.deb . && cp lib/x86_64-linux-gnu/* ."
```
danielezer, line 759: Didn't we say we are going to package the client in the docker image?

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -444,6 +444,10 @@ spec:
         - name: installer-info
           mountPath: "/artifactory_bootstrap/info/installer-info.json"
           subPath: installer-info.json
+      {{- if .Values.metrics.enabled }}
+        - name: artifactory-logs
+          mountPath: /opt/jfrog/artifactory/var/log
```
danielezer, line 449: Use `.Values.artifactory.persistence.mountPath` for the base path

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -533,6 +537,20 @@ spec:
 {{ toYaml .Values.filebeat.resources | indent 10 }}
         terminationGracePeriodSeconds: {{ .Values.terminationGracePeriod }}
       {{- end }}
+      {{- if .Values.metrics.enabled }}
```
danielezer, line 540: I don't understand the connection between `metrics.enabled` and `fluentd`. Why not introduce a separate flag for `fluentd`

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -533,6 +537,20 @@ spec:
...
+      - name: fluentd
+        image: "{{ .Values.metrics.fluentd.image.repository }}:{{ .Values.metrics.fluentd.image.tag }}"
+        volumeMounts:
+          - mountPath: /opt/bitnami/fluentd/conf/
+            name: fluentd-config
+          - mountPath: /opt/jfrog/artifactory/var/log
```
danielezer, line 546: Same, use `. Values.artifactory.persistence.mountPath`

stable/artifactory-ha/templates/metrics-service.yaml
```diff
@@ -0,0 +1,22 @@
...
+metadata:
+  name: {{ template "artifactory-ha.fullname" . }}-metrics
+  labels:
+    app: {{ template "artifactory-ha.name" . }}
+    chart: {{ template "artifactory-ha.chart" . }}
+    component: metrics
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+spec:
+  type: ClusterIP
```
danielezer, line 13: You should give the option to change the service type. Please use `.Values.metrics.service.type` here instead of hard-coding `ClusterIP`

stable/artifactory-ha/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,26 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: {{ template "artifactory-ha.name" . }}
+      component: metrics
+      release: {{ .Release.Name }}
+  endpoints:
+  - port: tcp-metrics
+    path: "/metrics"
+    {{- if .Values.metrics.serviceMonitor.interval }}
```
danielezer, line 20: Use `with` here instead of `if`

stable/artifactory-ha/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,26 @@
...
+    interval: {{ .Values.metrics.serviceMonitor.interval }}
+    {{- end }}
+    {{- if .Values.metrics.serviceMonitor.scrapeTimeout }}
```
danielezer, line 23: Same, please use `with`

danielezer (changes requested):

thamaraiselvam:
I have tested it on 
```
helm3
kubernetes 1.15
```

and it works fine for me

thamaraiselvam:
Should have option to add custom labels for `servicemonitor`  to support Prometheus operator

something like `artifactory-primary-statefulset.yaml`
```
{{- with .Values.metrics.serviceMonitor.labels }}
{{ toYaml . | indent 4 }}
{{- end }}
```

stable/artifactory-ha/README.md
rpf3: This diff is an artifact from my editor changing the line ending I think.  I'll fix it

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: My thinking with bundling them together was that the all of the JFrog documentation has the two tightly bundled and your support team indicated this was the recommended way to expose metrics.  If you think I should split them up, it shouldn't be too much work.

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: I think that is set to `/var/opt/jfrog/artifactory` which is slightly different and used for the artifact persistence not the log files.

https://github.com/jfrog/charts/blob/master/stable/artifactory-ha/values.yaml#L587

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
danielezer: `/var/opt/jfrog/artifactory/log/` also has the logs. I think its better to use that. Same for the `JFROG_HOME`. I don't think you need anything else other than the log folder, right?

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
danielezer: /var/opt/jfrog/artifactory/log/ also has the logs. I think its better to use that. Same for the JFROG_HOME. I don't think you need anything else other than the log folder, right?

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
danielezer: @eldada @chukka WDYT? Would appreciate your take here

stable/artifactory-ha/README.md
danielezer: Oh I see. Sorry about that

stable/artifactory-ha/templates/metrics-service.yaml
rpf3: fixed

stable/artifactory-ha/templates/servicemonitor.yaml
rpf3: fixed

stable/artifactory-ha/templates/servicemonitor.yaml
rpf3: fixed

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: No I don't, I'll make this change and thanks for the clarification

danielezer: 👍

peters95 (changes requested):
Please incorporate the changes on PR # 1097 to use fluent bit instead of fluentd. Once you have made the necessary changes to use fluent bit instead we can merge the two efforts together. You should be able to copy the fluent bit confs that are in that pull request as Artifactory and Artifactory HA use the same logs. We will need to update the sections for metrics though. Partner engineering can take on this effort to provide a valid prometheus fluent bit conf as a contribution to this effort.

rpf3:
> Please incorporate the changes on PR # 1097 to use fluent bit instead of fluentd. Once you have made the necessary changes to use fluent bit instead we can merge the two efforts together. You should be able to copy the fluent bit confs that are in that pull request as Artifactory and Artifactory HA use the same logs. We will need to update the sections for metrics though. Partner engineering can take on this effort to provide a valid prometheus fluent bit conf as a contribution to this effort.

@peters95 is there a Prometheus output plugin for Fluent-bit similar to the Fluentd one?  I have the Fluent-bit config running and parsing the log files but right now it's just configured to log to stdout.  Sorry, I think I'm just not familiar enough with this stack to know what the missing piece is here to bridge Fluent-bit to the ServiceMonitor.

https://github.com/fluent/fluent-plugin-prometheus

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: Pushed up a fix

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: all done

peters95:
@rpf3 yep no problem i think this is something partner engineering will have to take on. @jefferyfry can you look into level of effort to convert our prometheus fluentd to fluent bit. @rpf3 we will get back to you w/ a timeline. thanks!

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rimusz: In my opinion we should split them up, as not everyone will want to have fluentd if they just want prometheus metrics

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
```diff
@@ -668,6 +686,13 @@ spec:
         emptyDir:
           sizeLimit: {{ .Values.artifactory.persistence.size }}
   {{- end }}
+      {{- if .Values.metrics.enabled }}
+      - name: artifactory-logs
```
danielezer, line 690: What happens if you start off without `metrics.enabled` and decide you want to enable it later on? From when I understand, due to the conditional volume, all the logs will go away. I think this volume should not be conditioned and should always be created

stable/artifactory-ha/values.yaml
```diff
@@ -1570,3 +1570,366 @@ filebeat:
...
+    ## Timeout after which the scrape is ended
+    ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#endpoint
+    ##
+    scrapeTimeout: 15s
+
+  fluentd:
+    image:
+      ## The Bitnami Fluentd image has the Prometheus plugin pre-installed.
+      ##
+      repository: docker.io/bitnami/fluentd
```
danielezer, line 1602: Use `docker.bintray.io` please

stable/artifactory-ha/values.yaml
```diff
@@ -1570,3 +1570,366 @@ filebeat:
...
+            format_firstline /\d{4}-\d{1,2}-\d{1,2}/
+            format1 /^(?<timestamp>[^ ]*) \[(?<service_type>[^\]]*)\] \[(?<log_level>[^\]]*)\] \[(?<trace_id>[^\]]*)\] \[(?<class_line_number>.*)\] \[(?<thread>.*)\] - (?<message>.+)$/
+            time_key timestamp
+            time_format %Y-%m-%dT%H:%M:%S.%LZ
+        </parse>
+      </source>
+      <source>
+        @type tail
+        @id router_service_tail
+        path "#{ENV['JFROG_HOME']}/log/router-service.log"
```
danielezer, line 1704: Wouldn't it make more sense to set this environment variable to something like:
```
JFROG_LOGS_DIR={{ .Values.artifactory.persistence.mountPath }}/log
```
?
This would save the extra `/log`

danielezer (changes requested):

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: Generally I think what you are saying makes sense but I had a lot of back and forth with JFrog support about getting this working and was explicitly told, and your documentation seems to indicate this, that Fluentd was the supported and recommended way to expose Artifactory metrics.  I do agree that the chart should be flexible but I also think we should want to make it as frictionless as possible for developers to get metrics.  Maybe that means splitting them up and just making clearer documentation on this chart and your confluence pages

stable/artifactory-ha/values.yaml
rpf3: definitely a better idea, I'll push up the change

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: I'm not actually sure, I'll test it out now and confirm but I suspect you're correct.

rpf3: I don't think this is a problem because when you apply the Helm update to set `metrics.enabled` to true, the pods will get destroyed and recreated.  This is going to cause the volume to be destroyed as well since it is of type `emptyDir` which has the same lifetime of the pod.  I think a second confirmation would be good though.

stable/artifactory-ha/values.yaml
rpf3: I tried searching for `bitnami/fluentd` and couldn't find it, only a few other bitnami images pushed by either JFrog or VMware.  I also thought this would be fine because the Filebeat container is pulling from the Elastic Docker registry.  I'm fine with making this this change but did you want me to set up this image in Bintray?

danielezer: mmm I think it should be there, right @chukka ?

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
danielezer: Well, I think `emptyDir` is problematic. You want logs to last over time, and the lifespan of an `emptyDir` volume is the same as the pod's. For my other point, Consider a situation where you already have an Artifactory installation running, it already has some logs on the PVC, and now when you configure `metrics.enabled=true` the previously created logs will be overwritten by a new (empty) volume. That will be a problem. IMO you shouldn't use a volume specifically for logs and just use the main volume we already have. You can simply mount It in the fluentd container

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
danielezer: Yes, I think that's reasonable. It would be great to have `fluentd.enabled` and `metrics.enabled` separately

stable/artifactory-ha/values.yaml
danielezer: Please try to pull that image down. If it doesn't work, ignore my comment :)

rpf3: I tried with both `docker.bintray.io/bitnami/fluentd:latest` and `docker.bintray.io/bitnami/fluentd:1.11.2` and neither worked so I'm going to leave this for now.

danielezer: ok, sure. Thanks for trying

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: Pushed up a new commit decoupling these two.

stable/artifactory-ha/templates/artifactory-node-statefulset.yaml
rpf3: Pushed up a fixed version where I'm mounting the main volume instead.  Thanks for the comment.

stable/artifactory-ha/README.md
```diff
@@ -1244,6 +1244,18 @@ Then install nginx-ingress with the values file you created:
 helm upgrade --install nginx-ingress --namespace nginx-ingress center/kubernetes-ingress-nginx/ingress-nginx -f values.yaml
 ```
 
+### Prometheus Metrics
+
+If you want to enable Prometheus metrics you can use the `metrics` and `fluentd` configuration options.
```
danielezer, line 1249: Is fluentd required for Prometheus metrics? I'm not sure the coupling in the docs is justified. Maybe it should be a different section for each of them. WDYT?

stable/artifactory-ha/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ template "artifactory-ha.name" . }}
+      component: metrics
+      release: {{ .Release.Name }}
+  endpoints:
+  - port: tcp-metrics
+    path: "/metrics"
```
danielezer, line 22: I'm not very familiar with this object, but shouldn't this be in the values.yaml to allow changing it?

stable/artifactory-ha/values.yaml
```diff
@@ -1570,3 +1570,370 @@ filebeat:
...
+
+    ## Interval at which metrics should be scraped
+    ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#endpoint
+    ##
+    interval: 15s
+
+    ## Timeout after which the scrape is ended
+    ## ref: https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#endpoint
+    ##
+    scrapeTimeout: 15s
```
danielezer, line 1596: Should `scrapeTimeout` and `interval` really have the same value? Sounds like it might be problematic.

danielezer (commented):
I only have these few small comments. Other than that, it looks pretty good! Thanks for the contribution!

stable/artifactory-ha/README.md
rpf3: Yeah that's a good point, I'll break them out.

stable/artifactory-ha/templates/servicemonitor.yaml
rpf3: This is the default path for the Fluentd Prometheus output plugin.  We can make this configurable but we'd also need to use that same configuration value in the Fluentd configuration file.

https://github.com/fluent/fluent-plugin-prometheus

stable/artifactory-ha/values.yaml
rpf3: My thinking was that this would limit any individual scrape to the interval between them.  Which one would you suggest changing?

stable/artifactory-ha/templates/servicemonitor.yaml
danielezer: Currently, is it "hard-coded" in the fluetnd configuration as well? If so, I think it needs to move to the values.yaml rather than duplicating this value.

stable/artifactory-ha/values.yaml
danielezer: I think we can just leave it empty. WDYT? According to the docs I found, this is not mandatory. Just wrap each individual value with a `with` statement when you use it and it will act as an `if` statement

rpf3: That makes sense, leave it up to the consumers of the chart.

stable/artifactory-ha/templates/servicemonitor.yaml
rpf3: The port and path can be configured in Fluentd by editing this section of the configuration file:

```
# input plugin that exports metrics
<source>
  @type prometheus
  port 24231
  metrics_path /metrics
</source>
```

The best way I can think of to get configuration values from the Helm `values.yaml` into this Fluentd config file would be using environment variables similar to how we are setting the `JFROG_LOGS_DIR`.  Now that you bring this up I definitely think it's worth keeping them in sync and configurable.  Do you think this is a good way to do it?

danielezer: This config file uses `tpl` anyway, so you can simply use values as they are. e.g. `.Values.metrics.port`

rpf3:
@chukka can you provide some insight into the xray test that failed?  The link to the log returns a 404.

danielezer (approved):

chukka:
> @chukka can you provide some insight into the xray test that failed? The link to the log returns a 404.

Could you please rebase your changes and try again - If some thing fails I will look into it

rpf3:
@chukka I rebased but none of the checks have run in 18 hours which doesn’t seem normal. Am I missing something?

chukka:
@rpf3 Looks like CI builds are overloaded with multiple PR builds - checking - May take couple of hours to kick start this build

chukka:
@rpf3 Looks like build as started 👍

rpf3:
@chukka build succeeded, let me know if there is anything else I can do.

chukka:
@rpf3  Finally :)

@eldada @rimusz @rahulsadanandan  Can you please review once , This has been pending for long time

peters95:
We are discussing this PR a bit later today but the overall idea is this makes sense to merge and then later we can support fluent bit.

peters95 (changes requested):
Please use the standard fluentd confs in the jfrog/log-analytics folder. This way customers who do not want to use prometheus could swap the confs. Please use the env JFROG_HOME instead of JFROG_LOG_DIRs. This way people could disable metrics and swap the confs out for splunk,datadog, or elastic as well.

rpf3:
@peters95 just for clarity, you would like me to change the default Fluentd config to be this file:

https://github.com/jfrog/log-analytics/blob/master/fluentd/fluent.conf.rt

My concern is that these configs in your repo are incorrect.  I worked with your support team a few weeks ago and the pointed me to this repo where I then opened this PR to fix the ordering of the request/response parsing.  Your support team also told me that they would update your Wiki docs because the ordering was incorrect there as well.

https://github.com/jefferyfry/log-analytics/pull/1/files

cc @jefferyfry

peters95:
Please use the official repo not jeff's personal fork. Here is the link:

https://github.com/jfrog/log-analytics-prometheus

Submit your changes to this repo and I will make sure they get merged so we can all use the official version. This will enable us to also support all the vendors.

Again much appreciated though for all your hard work I can tell you put a lot of effort and its very appreciated.

rpf3:
@peters95 it actually looks like your new Prometheus repo is correct but some of the others aren't.  Here is the PR for the base log-analytics and I'll submit the vendor ones later today.

https://github.com/jfrog/log-analytics/pull/40

Changes to the PR are incoming

rpf3:
Can you add some clarity around the env var change?  I worked with @danielezer on that as I had originally hardcoded a path to the JFROG_HOME variable but using the `persistence.mountPath` seemed cleaner since then it would follow the log volume.

peters95:
@rpf3 the JFROG_HOME was chosen for it to work in both k8s and non-k8s environments. Generally speaking we ask the users of either installation to setup /opt/jfrog or whatever directory they plan to install the JFROG products into. This is keeping with the regular installation procedures for non-k8s to keep things similar for our customers who use both.

rpf3:
For reference these are the two commits that brought me to the current state:

https://github.com/jfrog/charts/pull/1094/commits/8154fe6c220dcd6bdffbd6e28966fcb503dfd2bf
https://github.com/jfrog/charts/pull/1094/commits/c68decf2b04f29d10e7754722dd034e0d01cc964

danielezer:
> @rpf3 the JFROG_HOME was chosen for it to work in both k8s and non-k8s environments. Generally speaking we ask the users of non-k8s installation to setup /opt/jfrog or whatever directory they plan to install the JFROG products into. This is keeping with the regular installation procedures for non-k8s to keep things similar for our customers who use both.
> 
> I see your point though about the mount point. Let me noodle on this for a few minutes and get back to you with an answer.

@peters95 The JFROG_LOGS_DIR is not used in a JFrog product. There's no need to follow the convention for JFrog installation. This is only used in the fluentd container

peters95:
@danielezer  The issue is our fluentd confs need to work in both non-k8s or k8s. We are discussing if it makes more sense to request for non-k8s to use a similar env JFROG_LOGS_DIR to unify the confs between the two. Ideally we would like to only support 1 set of fluentd confs.

peters95:
JF_PRODUCT_HOME is defined by default in the docker image and it points to: /opt/jfrog/artifactory

We will update our fluent confs to use this environment instead as it eliminate the step of the user defining JFROG_HOME to /opt/jfrog.

@rpf3 can you update your confs to use this as well. No need to add the extra ENV based off the mount point as its defined already in the base image.

rpf3:
We'll still need to add the environment variable to the Fluentd container but if we configure that to match the Artifactory value we just need to prefix all the log paths in the Flunetd conf to be `#{ENV['JF_PRODUCT_HOME']}/var/log/*.log`

peters95:
We are updating all the fluentd confs to use JF_PRODUCT_HOME. The only exception is pipelines which has LOG_DIR. We submitted a JIRA to pipelines team to update their docker images to include this ENV as all the other product images do.

danielezer:
Are you planning to duplicate the config from https://github.com/jfrog/log-analytics/pull/40/files ? I don't think it really makes sense. It's not feasible to manage these two big configuration files. You should just download the files with an init container or something like that

peters95:
@rpf3 I will update your branch to use the new ENV. Note we will need you to sign the CLA tho once https://github.com/jfrog/charts/pull/1143 is merged to be able to merge this PR.

peters95:
@danielezer  that makes sense to me. I think we should do this in a follow up PR tho where I will add support for Splunk, Datadog, and Elastic.

peters95:
@rpf3 i have updated your branch w/ the changes. Once https://github.com/jfrog/charts/pull/1143 is merged I request you sign the CLA so we can merge this PR.

peters95 (approved):

rpf3:
@peters95 I don't think the config you added is going to work, the `JF_PRODUCT_HOME` env var only exists in the Artifactory container, not the Fluentd container.  I think we need to add `JF_PRODUCT_HOME` to Fluentd as well as change the `artifactory-volume` mount path on the Fluentd container.

I've done this locally by setting the var and mount path both to `/opt/jfrog/artifactory` and then updating the config to be of the form `#{ENV['JF_PRODUCT_HOME']}/log/artifactory-service.log`

peters95:
@rpf3 good point thanks the only change we need is to add var to the path for it to be:

#{ENV['JF_PRODUCT_HOME']}/var/log/artifactory-service.log

Ill update the env to include this thanks

danielezer:
If that's the case I would revisit the JFROG_LOGS_DIR env var.

peters95:
@rpf3 added back the env as           
          - name: JF_PRODUCT_HOME
            value: /opt/jfrog/artifactory

danielezer:
You don't really avoid providing any configuration, so let's just use the simplest solution and the one that requires the least code duplication, which is the `JFROG_LOGS_DIR`

peters95:
@danielezer I think JF_PRODUCT_HOME still makes sense as it will work ootb for customers who use docker compose. We have a far amount of customers who setup their own docker containers and do not use our helm charts.

danielezer:
It won't work OOTB because you would have to provide the `JF_PRODUCT_HOME` to the fluentd container. If you set an environment variable anyway, why not use the one that requires the least duplication?

peters95:
@danielezer We provide a zip installer and expect customers to untar and run fluentd directly as the artifactory user not as a sidecar. sidecar setup was only added in helm.

danielezer:
Are you sure? That implies that the Docker image is different between the helm chart and the docker-compose. that sounds off to me...

rpf3:
I still think we're missing what I mentioned above.  The `artifactory-volume` where the logs are is going to be mounted to `/var/opt/jfrog/artifactory/log` but Fluentd is going to be looking for the log files at `/opt/jfrog/artifactory/var/*.log`.

My suggestion is that we change the mountPath on the Fluentd container to `/opt/jfrog/artifactory` and change the config from `JF_PRODUCT_HOME/var` to `JF_PRODUCT_HOME/log`

peters95:
@danielezer should we add this into values.yaml instead of hard code. I agree w/ the duplication comment. For example in the pipelines values.yaml they define both:

  mountPath: /opt/jfrog/pipelines/var/etc

  logPath: /opt/jfrog/pipelines/var/log

danielezer:
@rpf3 Does `JF_PRODUCT_HOME` point to `/var/opt...` or to `/opt/....` ? I would like to avoid duplication as much as we can. If they are different, we can just set `JF_PRODUCT_HOME` to `/var/opt....`, this will only be applicable for the fluentd container anyway.

@peters95 Yes, I think that makes sense.

peters95:
@rpf3 @danielezer I will add the logPath as a new value in values.yaml and update the ENV to point to this instead.

rpf3:
@danielezer yeah if we set `JF_PRODUCT_HOME` to `{{ .Values.artifactory.persistence.mountPath }}` then we still need to update the fluentd config to be `JF_PRODUCT_HOME/log`

danielezer:
@rpf3 Yes, that's true. But if it is set in `values.yaml` it's not a big deal

peters95:
I added a productPath to values.yaml and then this should work I believe in either case:

image

danielezer:
I don't like that name. Just point it to `{{ .Values.artifactory.persistence.mountPath }}`. I want to avoid using both `/var/opt/...` and `/var/....`. Lets just use `{{ .Values.artifactory.persistence.mountPath }}` and set `JF_PRODUCT_HOME` to `{{ .Values.artifactory.persistence.mountPath }}` in the fluentd container

rpf3:
Maybe I've been looking at this for too long.  If you do the above then the volume will be mounted to `/opt/jfrog/artifactory/var` which means the log files will be in `/opt/jfrog/artifactory/var/log/*.log` but your Fluentd changes are still looking for `JF_PRODUCT_HOME/var/*.log`

peters95:
@rpf3 /opt/jfrog/artifactory/var/log/*.log is the correct folder

danielezer:
@peters95 nope, let's use `/var/opt/jfrog/artifactory`, as it is set in `{{ .Values.artifactory.persistence.mountPath }}`

peters95:
@danielezer ok but I thought the plain was to init container and pull the confs from the official repo long term. this will make it not work w/o modifications which we will have to keep two copies then.

danielezer:
@peters95 we will still be able to use the one from the official repo. What I'm saying is to use `JFROG_PRODUCT_HOME/log` in the fluentd config file, as you suggested. In the fulentd container, where we control the value of `JF_PRODUCT_HOME` you can just set it to `/var/opt/jfrog/artifactory` which is already in the `values.yaml`. `/var/opt/jfrog/artifactory/log` and `/opt/jfrog/artifactory/var/log` are the same. One is a soft link to the other.

peters95:
@danielezer makes sense thanks will update and you can review.

peters95:
@danielezer this works for everything except pipelines. Also there is already an ENV defined for this in docker images:

JF_PRODUCT_DATA_INTERNAL=/var/opt/jfrog/artifactory

We should use this name instead as it will work for both docker compose or helm then.

danielezer:
Cool, sounds good

peters95:
@rpf3 @danielezer I've made the changes but looks like there may be some other issues in these charts as it fails to install using the below command:

helm install artifactory-ha . --set nginx.service.ssloffload=true --set nginx.tlsSecretName=tls-ingress --set artifactory.node.replicaCount=1 --set artifactory.masterKey=FFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF --set artifactory.joinKey=EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE  --set artifactory.license.secret=artifactory-license --set artifactory.license.dataKey=artifactory.cluster.license --set metrics.enabled=true --set fluentd.enabled=true

Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"

If i drop the --set metrics.enabled=true  it works.

peters95 (changes requested):
Adding the --set metrics.enabled=true results in error during installation:

Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"

peters95:
The volume mount also does not appear to be properly configured on the sidecar.. statefulsets hangs with:

Warning  FailedCreate      47s (x15 over 2m10s)  statefulset-controller  create Pod artifactory-ha-artifactory-ha-primary-0 in StatefulSet artifactory-ha-artifactory-ha-primary failed error: Pod "artifactory-ha-artifactory-ha-primary-0" is invalid: spec.containers[1].volumeMounts[1].name: Not found: "artifactory-volume"

rpf3:
The servicemonitor error is because it's expecting prometheus operator to already be installed in the cluster already.  Maybe we need to either add that into the README or add an option to create the CRD

peters95:
@rpf3 can you update the readme to include the operator installation. I will continue testing the charts and then commit these changes.

peters95:
Thanks @rpf3 for the readme update.

I noticed some errors in the prometheus conf.. not related to anything you put @rpf3 .. I'm patching the source confs in the official repo and updating the confs in the helm chart to match. For example service logs were not properly being delivered:

2020-09-24 21:19:48 +0000 [warn]: #0 [frontend_service_tail] pattern not matched: "2020-09-24T21:19:46.790Z \e[33m[jffe ]\e[39m \e[34m[INFO ]\e[39m [                ] [                              ] [main                ] - Recurring tasks started"

Give me an hour or so to make all these changes and test everything end to end.

peters95:
Hi @rpf3 

We would like to merge, but it looks like we had a problem with our CLA Assistant bot.
We would really appreciate it if you could sign our [Contributor License Agreement](https://gist.github.com/jfrog-ecosystem/7d4fbeaac18edbd3cfc38831125acbb3) before we can accept your contribution. You can sign the CLA by just posting a Pull Request Comment same as the below format.

Thanks again!

peters95 (approved):

peters95:
@rpf3 Please comment on this PR that you have viewed and agreed to the CLA so we can merge this PR.

peters95:
@rpf3 We won't be able to merge this without your acceptance of the CLA. Please advise if you wish to close this PR instead.

rpf3:
@peters95 sorry, since I was working on this for work I had to send that CLA to our legal team to review.  I'll try and get back to you ASAP as I definitely want this merged.

rpf3:

peters95:
LGTM. Tests were run in partner engineering to confirm this works for prometheus.

rpf3:
@peters95 can you help me troubleshoot this CLA error?

peters95:

peters95:
recheckcla

peters95:
@rpf3 try and leave two new comments like I did above and see if that resolves the issue. apologizes we have been making updates to our CLA recently.

rpf3:

rpf3:
recheckcla

rpf3:
recheckcla

peters95:
@rpf3 I've reached out to the team here who has added the CLA to look into the above issue. @eyalbe4 FYI looks like CLA is not approving even though @rpf3 has commented he agrees and recheckcla was run.

peters95:
Spoke w/ our CLA team they said it could be for various reasons such as spaces that they are working on. For now they will add @rpf3 manually as he has commented acceptance and we can merge this PR.

jefferyfry (approved):
LGTM

rahulsadanandan:
Hi @rpf3 Unfortunately, we had to revert these changes in the latest release of chart #1319 
This was done following an internal discussion, not to have the log analytics resources included in our core helm chart.
The alternative working solution is added as part of another repository
https://github.com/jfrog/log-analytics-prometheus
We have added a working example there
https://github.com/jfrog/log-analytics-prometheus/blob/master/artifactory-values.yaml
cc @eldada @peters95 @chukka @JfrogPrasanna

rpf3:
@rahulsadanandan thanks for the update

PR #1109, opened 2020-09-16
[artifactory-ha] update nginx TLS config

dhutty:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

As discussed with billm@jfrog.com, this brings the nginx TLS config into line with the "Intermediate" recommendations from https://ssl-config.mozilla.org

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [4.0.3] - Sep 16, 2020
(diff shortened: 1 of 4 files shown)


eldada:
Artifactory has many other, older clients that use it. Much more than web browsers.
Some of these are fairly old, and we try and keep support for a fairly wide range of clients.
I'm worried that this change might actually break this.
I suggest you wrap this configuration in an `if` clause with an option to select turning it on or off.
I'll also take it internally to see if we can verify this somehow.

danielezer (approved):

gitta-jfrog:
We decided not to take this change. In any case, customers can pass their own custom "artifactoryConf". Thanks!

Issue #859, opened 2020-04-26
[artifactory] Database credentials are not updated correctly when artifactory is recreated.


hetii:
Hi.

I play with a helmfile that setup for me artifactory-oss.
As soon as I run it second time the password for the database is change and artifactory is not able to use it again.
This lead to readiness failure and the only way to fix it is to delete pvc of artifactory-volume-artifactory-artifactory-0 :< 
This issue is related to https://github.com/jfrog/charts/issues/63

**Version of Helm and Kubernetes**: 
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}
Kubernetes: v1.15.10

**Which chart**:  2.2.0 

**What happened**:
Database password is upgraded and other component are no aware about this fact.

**What you expected to happen**:
All components used PostgreSQL db should be aware of db password changes.

**How to reproduce it** (as minimally and precisely as possible):
Install few times char in version 2.2.0.

chukka:
@hetii Can you share the helm command used to install the chart initially (is postgresql password passed as part of values)  so that i reproduce the issue 

BTW - after the first install, for upgrades you need to pass the database password so a new one will not be generated that might conflict with previous one.

chukka:
@hetii  is this still an issue ? waiting for your update

kataliyst:
I have this same issue.  I deploy the chart with Rancher.  Any upgrades stall and fail.  rollback doesn't work.  logs show this password problem is the issue - passwords are auto-generated.  First deployment is fine - but as I say any upgrade or rollback results in this issue.  This is artifactory-oss 2.2.2 -> 2.2.3:

DETAIL:  Password does not match for user "artifactory". 
	Connection matched pg_hba.conf line 1: "host     all             all             0.0.0.0/0               md5" 
FATAL:  password authentication failed for user "artifactory" 
DETAIL:  Password does not match for user "artifactory".

chukka:
@hetii  @kataliyst  can you share your values.yaml file and steps to reproduce ?

kataliyst:
As I mentioned, I use Rancher.  Simply add the Jfrog charts to Rancher's catalogs.  Go into the application screen to deploy an app.  Launch.  Select Artifactory-oss and deploy.  If it's a new install, it will be up in a minute.  If you attempt to upgrade this to a new version, it will fail and you will get complaints about the password.

[...]

PR #1110, opened 2020-09-16
[artifactory] Create database secrets if postgres password is set

woodcockjosh:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

Allows user to set database password rather than use the generated one. This also works as a workaround for https://github.com/jfrog/charts/issues/859

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes 

### DIFF
--- a/stable/artifactory/templates/artifactory-database-secrets.yaml
+++ b/stable/artifactory/templates/artifactory-database-secrets.yaml
@@ -1,4 +1,4 @@
-{{- if and  (not .Values.database.secrets) (not .Values.postgresql.enabled) }}
+{{- if or (and  (not .Values.database.secrets) (not .Values.postgresql.enabled)) .Values.database.password }}
 apiVersion: v1
 kind: Secret
 metadata:
--- a/stable/artifactory/values.yaml
+++ b/stable/artifactory/values.yaml
@@ -429,6 +429,9 @@ artifactory:
         url: "jdbc:postgresql://{{ .Release.Name }}-postgresql:{{ .Values.postgresql.service.port }}/{{ .Values.postgresql.postgresqlDatabase }}"
         driver: org.postgresql.Driver
         username: "{{ .Values.postgresql.postgresqlUsername }}"
+        {{- if .Values.postgresql.postgresqlPassword }}
+        password: "{{ .Values.postgresql.postgresqlPassword }}"
+        {{- end }}
         {{- else }}
         type: "{{ .Values.database.type }}"
         driver: "{{ .Values.database.driver }}"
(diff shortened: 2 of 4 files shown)


danielezer (approved):

stable/artifactory/templates/artifactory-database-secrets.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and  (not .Values.database.secrets) (not .Values.postgresql.enabled) }}
+{{- if or (and  (not .Values.database.secrets) (not .Values.postgresql.enabled)) .Values.database.password }}
```
chukka, line 1: @woodcockjosh I'm not sure this change is required , `.Values.database.password` implies external postgres

stable/artifactory/values.yaml
```diff
@@ -429,6 +429,9 @@ artifactory:
         url: "jdbc:postgresql://{{ .Release.Name }}-postgresql:{{ .Values.postgresql.service.port }}/{{ .Values.postgresql.postgresqlDatabase }}"
         driver: org.postgresql.Driver
         username: "{{ .Values.postgresql.postgresqlUsername }}"
+        {{- if .Values.postgresql.postgresqlPassword }}
+        password: "{{ .Values.postgresql.postgresqlPassword }}"
```
chukka, line 433: @woodcockjosh [This](https://github.com/helm/charts/blob/master/stable/postgresql/templates/secrets.yaml#L16) creates a secret for `.Values.postgresql.postgresqlPassword` from postgresql sub-chart and in statefulset it is referenced using **JF_SHARED_DATABASE_PASSWORD**
```
{{ if or .Values.database.secrets.password .Values.database.password .Values.postgresql.enabled }}
        - name: JF_SHARED_DATABASE_PASSWORD
          valueFrom:
            secretKeyRef:
        {{- if .Values.database.secrets.password }}
              name: {{ tpl .Values.database.secrets.password.name . }}
              key: {{ tpl .Values.database.secrets.password.key . }}
        {{- else if .Values.database.password }}
              name: {{ template "artifactory.fullname" . }}-database-creds
              key: db-password
        {{- else if .Values.postgresql.enabled }}
              name: {{ .Release.Name }}-postgresql
              key: postgresql-password
        {{- end }}
      {{- end }}
```

woodcockjosh: right well until a solution is implemented for #859 I'm going to use this approach to fix the problem with upgrades. It would be a shame if everyone else had to copy paste the artifactory chart to solve their problem but I understand if upgrading the chart without the upgrade failing is not the highest priority for the repo.

woodcockjosh:
@chukka any update on this?

gitta-jfrog:
We decided to keep the current implementation as it is, if needed, please raise a issue and we will investigate. Thanks!

PR #1170, opened 2020-10-01
[XRAY] Router user id and group id support in values.yaml

peters95:
#### PR Checklist
- [X] Chart Version bumped
- [X] CHANGELOG.md updated
- [X] Title of the PR starts with chart name (e.g. `[artifactory]`)

Enables support for running all xray containers in a custom user id and/or group id range. Currently we support it in 4 out of the 5 containers in the xray pod through common.xrayUserId and common.xrayGroupId in values.yaml so this adds similar functionality but for the router container.

PTRENG-1120 required for Openshift to enable all containers to use a custom user and group id range.

Required for Openshift for customer fixes.

### DIFF
--- a/stable/xray/values.yaml
+++ b/stable/xray/values.yaml
@@ -367,6 +367,8 @@ common:
   # xrayVersion:
   xrayUserId: 1035
   xrayGroupId: 1035
+  routerUserId: 1117
+  routerGroupId: 1117
 
   # Xray configuration to be written to xray_config.yaml
   xrayConfig:
--- a/stable/xray/CHANGELOG.md
+++ b/stable/xray/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Xray Chart Changelog
 All changes to this chart will be documented in this file.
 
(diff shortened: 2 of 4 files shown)


jefferyfry (approved):
LGTM

stable/xray/values.yaml
```diff
@@ -367,6 +367,8 @@ common:
   # xrayVersion:
   xrayUserId: 1035
   xrayGroupId: 1035
+  routerUserId: 1117
```
danielezer, line 370: I think this belongs in the router section of the values.yaml, i.e. .Values.router.userId and .Values.router.groupId
The common section is intended for values that are common for all of xray’s micro services

danielezer (changes requested):

vinayagg (commented):
Daniel has good point, other than that, LGTM

peters95:
The spec overrides the router container image. Normally router uses a different user 1117 vs 1035 but we can change our Openshift images instead to match this configuration instead of making changes to xray base chart.

PR #1260, opened 2020-11-13
[artifactory/ artifactory-ha] Update check for making tls keypair to use nginx.https.enabled

### LINKED ISSUE #1259: Chart makes tls.crt and tls.key even when Values.nginx.https.enabled = false
BUG REPORT

**Version of Helm and Kubernetes**: 2.x, 1.13

**Which chart**: artifactory-ha but also artifactory

**What happened**: The generated chart makes a tls.key and tls.crt when Values.nginx.https.enabled is set to false

**What you expected to happen**: It should not make those files

**How to reproduce it** (as minimally and precisely as possible): helm upgrade --install artifactory-ha --dry-run --debug --set nginx.https.enabled=false center/jfrog/artifactory-ha 

**Anything else we need to know**: pr incoming

anderm3:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

**What this PR does / why we need it**: This PR updates the logic around making tls.key and tls.crt files so they are not made when nginx.https.enabled is false.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #1259

I can add an additional check for nginx.enabled too if you would like but it looked a little gross when I wrote it initially which is why I left it out on this PR.

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [4.4.6] - Nov 13, 2020
+* Fix issue with tls file generation when nginx.https.enabled is false
+
 ## [4.4.5] - Nov 9, 2020
 * Updated Artifactory version to 7.10.6 - [Release Notes](https://www.jfrog.com/confluence/display/JFROG/Artifactory+Release+Notes#ArtifactoryReleaseNotes-Artifactory7.10.6)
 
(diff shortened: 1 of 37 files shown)


anderm3:

rahulsadanandan:
@anderm3 
Thanks for your contribution. Would you mind adding the logic in `nginx-deployment.yaml` as well to handle the volumes and volumeMounts for ssl-certificates . Otherwise it would fail with `volume "ssl-certificates" : secret not found` when `nginx.https.enabled` is set to false

cc @chukka

chukka (approved):

rahulsadanandan (approved):
LGTM

chukka (approved):

PR #1329, opened 2021-01-05
[artifactory-ha] stateful set spec.serviceName should match the service created

kpirateone:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

currently the service generated from artifactory service creates a service with name: `artifactory-ha.fullname`.  the artifactory-node-statefulset.yaml creates a spec.serviceName as `artifactory-ha.node.name`.  the service specified in the statefulset needs to align, particularly when addressing pods directly in kubernetes DNS : https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

this pr changes the service creation to have the same name as specified in the node statefulset spec.serviceName

none

@eldada @rimusz @danielezer

### DIFF
--- a/stable/artifactory-ha/CHANGELOG.md
+++ b/stable/artifactory-ha/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory-ha Chart Changelog
 All changes to this chart will be documented in this file
 
+## [4.7.5] - Jan 05, 2020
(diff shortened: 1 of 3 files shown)


gitta-jfrog:
We decided not to implement this change as this might have impact on our current deployments.
In addition, the node sts should not be used anymore as only primary sts is in use.

PR #1430, opened 2021-05-09
[artifactory][artifactory-ha] Allow overriding the urlStrictPolicy setting for webhooks

KierranM:
#### PR Checklist
- [ ] Chart Version bumped
- [x] CHANGELOG.md updated

This allows setting the `event.webhooks.urlStrictPolicy` setting, this is needed in environments where the webhook target is private. Without this you are unable to save a new webhook pointing at an internal service and the logs throw this error:
```
error validating webhook url: illegal private network: <internal ip here> [subscriptions]
```

I wasn't sure how to (or whether to) do the version bump since the changes to align the chart version with the app version.

### DIFF
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -253,6 +253,11 @@ artifactory:
   openMetrics:
     enabled: false
 
+  webhooks:
+    ## when true private networks (loopback, RFC1918, RFC3927, and IPV6 unique local addresses) will not be allowed as
+    ## webhook targets
+    urlStrictPolicy: true
+
   # This directory is intended for use with NFS eventual configuration for HA
   haDataDir:
     enabled: false
@@ -610,6 +615,9 @@ artifactory:
     replicator:
       enabled: true
     {{- end }}
+    event:
+      webhooks:
+        urlStrictPolicy: {{ .Values.artifactory.webhooks.urlStrictPolicy }}
 
   ## IMPORTANT: If overriding artifactory.internalPort:
   ## DO NOT use port lower than 1024 as Artifactory runs as non-root and cannot bind to ports lower than 1024!
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -253,7 +253,7 @@ artifactory:
   openMetrics:
     enabled: false
 
-  events:
+  event:
     webhooks:
       ## when true private networks (loopback, RFC1918, RFC3927, and IPV6 unique local addresses) will not be allowed as
       ## webhook targets
@@ -618,7 +618,7 @@ artifactory:
     {{- end }}
     event:
       webhooks:
-        urlStrictPolicy: {{ .Values.artifactory.events.webhooks.urlStrictPolicy }}
+        urlStrictPolicy: {{ .Values.artifactory.event.webhooks.urlStrictPolicy }}
 
   ## IMPORTANT: If overriding artifactory.internalPort:
   ## DO NOT use port lower than 1024 as Artifactory runs as non-root and cannot bind to ports lower than 1024!
--- a/stable/artifactory-ha/values.yaml
+++ b/stable/artifactory-ha/values.yaml
@@ -253,10 +253,11 @@ artifactory:
   openMetrics:
     enabled: false
 
-  webhooks:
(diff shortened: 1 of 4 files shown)


chukka:
@KierranM  Thanks for this contribution , we would like to review this internally and take it forward from there, This change will be part of 7.19.x release. Note: No further is action is needed from you .

stable/artifactory-ha/values.yaml
```diff
@@ -253,6 +253,11 @@ artifactory:
   openMetrics:
     enabled: false
 
+  webhooks:
```
chukka, line 256: @KierranM  please change this to events.webhooks.urlStrictPolicy : true

stable/artifactory-ha/values.yaml
```diff
@@ -610,6 +615,9 @@ artifactory:
     replicator:
       enabled: true
     {{- end }}
+    event:
+      webhooks:
+        urlStrictPolicy: {{ .Values.artifactory.webhooks.urlStrictPolicy }}
```
chukka, line 620: please change this to .Values.artifactory.events.webhooks.urlStrictPolicy

stable/artifactory/values.yaml
```diff
@@ -539,6 +544,9 @@ artifactory:
     replicator:
       enabled: true
     {{- end }}
+    event:
+      webhooks:
+        urlStrictPolicy: {{ .Values.artifactory.webhooks.urlStrictPolicy }}
```
chukka, line 549: please change this to .Values.artifactory.events.webhooks.urlStrictPolicy

stable/artifactory/values.yaml
```diff
@@ -209,6 +209,11 @@ artifactory:
   openMetrics:
     enabled: false
 
+  webhooks:
```
chukka, line 212: please change this to .events.webhooks.urlStrictPolicy: true

chukka (commented):
@KierranM Thanks for this contribution , Can you address these review comments, Post that we would like to review this internally and take it forward from there, This change will be part of 7.19.x release. Note: No further action will be  needed from you post these review comments

chukka (changes requested):
@KierranM just verified  with https://www.jfrog.com/confluence/display/JFROG/Artifactory+System+YAML , event should be root element (not under artifactory), Can you make corresponding changes

chukka:
@KierranM Also , thinking if we can set this to default false ? customers can enable by setting this to true as needed.

Also, looking at systemyaml configuration, by default it was true, was it not working with previous charts with adding any configuration ?

chukka:
@KierranM Can you please verify if it works without this PR (as default configuration is to allow `urlStrictPolicy` to true)

KierranM:
> @KierranM just verified with https://www.jfrog.com/confluence/display/JFROG/Artifactory+System+YAML , event should be root element (not under artifactory), Can you make corresponding changes

I'm currently away from a computer for the next few days, but I'll update this as soon as I'm back

KierranM:
> @KierranM Can you please verify if it works without this PR (as default configuration is to allow `urlStrictPolicy` to true)

The issue we're facing is _because_ it defaults to `true`. We need to be able to disable this setting otherwise we can't create webhooks pointing at internal APIs

chukka:
> > @KierranM just verified with https://www.jfrog.com/confluence/display/JFROG/Artifactory+System+YAML , event should be root element (not under artifactory), Can you make corresponding changes
> 
> I'm currently away from a computer for the next few days, but I'll update this as soon as I'm back

Sure

chukka:
@KierranM If you would like to modify default systemyaml (defined in values.yaml), you can make use of [systemYamlOverride](https://github.com/jfrog/charts/blob/master/stable/artifactory/values.yaml#L66) feature .I'm closing this

KierranM:
@chukka, it's true we _can_ use the override, but you can only override the _entire_ system yaml. Doing it this way means we have to check the source chart every time we do an upgrade (either chart version or artifactory version) and make sure we copy over any changes that have been made here. I'm concerned about the amount of potential for errors this manual process introduces.

Is there a particular reason you aren't willing to allow this to be configurable as a chart parameter?

chukka:
@KierranM I think `systemYamlOverride` approach is for advanced users. However , Its only one time configuration change and doesn't impact with chart changes going forward, you know what values can be in system.yaml . I prefer this approach  as you have complete control over systemyaml changes.

Our current systemyaml is already very lengthy and wouldn't want to overload with additional changes.

PR #1858, opened 2024-02-02
[artifactory] Add the ability to set `signedUrlExpirySeconds` to google providers

heyitsmdr:
#### PR Checklist
- [x] Chart Version bumped
- [x] CHANGELOG.md updated

We are using the `google-storage-v2-direct` provider and were previously providing a custom binary store with a `signedUrlExpirySeconds` set to `120`. We recently switched to the chart-provided binary store config and can no longer control the signed URL expiry. As such, we are now using the provider default of `30`.

We've recently ran into issues where we would like to increase the expiry a bit, and we're not currently able to due to the binary store template provided by the chart not providing this ability. We would rather not go back to a custom-defined binary store config, so I'm raising this PR to address that.

I made `signedUrlExpirySeconds` conditional on the value being set in a Helm deployment. I was going to add `.artifactory.persistence.googleStorage.signedUrlExpirySeconds: 30` to [/stable/artifactory/values.yaml](https://github.com/jfrog/charts/blob/master/stable/artifactory/values.yaml#L941), and remove the conditional from the binary store template, but figured this approach would be better so that, if not set, we can just rely on the built-in provider default without having to manage it within this Helm chart.

### DIFF
--- a/stable/artifactory/CHANGELOG.md
+++ b/stable/artifactory/CHANGELOG.md
@@ -1,6 +1,9 @@
 # JFrog Artifactory Chart Changelog
 All changes to this chart will be documented in this file.
 
+## [107.77.4] - Feb 02, 2024
+* Added `signedUrlExpirySeconds` option to artifactory.persistence.type of `google-storage`, `google-storage-v2`, and `google-storage-v2-direct`
+
 ## [107.77.3] - Jan 16, 2024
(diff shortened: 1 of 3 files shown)


heyitsmdr:

heyitsmdr:
recheckcla

PR #1862, opened 2024-02-16
[artifactory-ha] Add Annotations to Primary Artifactory Service and Resources to Nginx Deployment

yellowplane13:
Add Annotations to primary artifactory `service` 
Add resources to `setup` container in the nginx deployment yaml

#### PR Checklist

Add Annotations to Primary Artifactory Service: This change adds annotations to the primary Artifactory service. 

Add resources block to setup container in nginx deployment yaml: This change gives the ability to change limits and requests for the `setup` init container within the nginx deployment.


yellowplane13:

stable/artifactory-ha/templates/nginx-deployment.yaml
```diff
@@ -79,6 +79,8 @@ spec:
         volumeMounts:
         - mountPath: {{ .Values.nginx.persistence.mountPath | quote }}
           name: nginx-volume
+        resources:
+{{ toYaml .Values.nginx.resources | indent 10 }}
```
se-kai, line 83: maybe "{{ toYaml .Values.initContainers.resources | indent 10 }}" better? this is init container

yellowplane13: Yes, you're right. Updated it

yellowplane13: Hi @se-kai , could you please review the PR again? 
Tagging @oumkale and @chukka as well. As soon as this is merged into a release, we can use this as our upstream chart instead of downloading the entire thing into our repo. Thank you 🙏

se-kai: this is already fixed and avialble in current main branch
https://github.com/jfrog/charts/blob/master/stable/artifactory-ha/templates/nginx-deployment.yaml#L79
I used it already and it works fine for me.

yellowplane13: Hi @se-kai , thank you for the update. Appreciate having resources block added to nginx-deployment. 

We also need `annotations` added to `artifactory-primary-service.yaml`'s `metadata`. It is not originally added and we can't use this chart as an upstream until that's added unfortunately. Can that be taken into account for the next release? 
I can rebase and update this PR with just the `artifactory-primary-service.yaml` changes required if you need.

stable/artifactory-ha/templates/nginx-deployment.yaml
```diff
@@ -81,6 +81,8 @@ spec:
         volumeMounts:
         - mountPath: {{ .Values.nginx.persistence.mountPath | quote }}
           name: nginx-volume
+        resources:
+{{ toYaml .Values.initContainers.resources | indent 10 }}
```
oumkale, line 85: `resources` have already been mentioned at line number 79. [Ref](https://github.com/jfrog/charts/pull/1862/files#diff-7e14e678a545b74eb7cca035f06da7254c66b9cf7cab22692d631322571a17ceR79)

PR #2154, opened 2026-01-05
[catalog] Fix resources for router container in catalog

timmy59100:
resources in router container were only applied when containerSecurityContext was enabled.


timmy59100:
recheck

megha1906 (approved):

PR #2522, opened 2026-08-31
artifactory: split PEM certificate bundles in customCertificates init container

### LINKED ISSUE #2419: `global.customCertificates` should support bundles
Hi,

Currently the key `global.customCertificates` expects one secret containing one or more CA, with one certificate per key.
However, people might manage trust CA bundles as one unique key containing all certificates (i.e. with trust-manager), which is not working as expected on artifactory side : a truststore entry is added but from the whole bundle, so certificates are not trusted.

Maybe an extra checking step on the initcontainer would be sufficient ?

amarkdotdev:
Fixes #2419

When a customCertificates secret stores multiple CAs in one PEM bundle (for example from trust-manager), the init container now splits the bundle into individual certificate files before copying them into the truststore.

## Test plan
- [x] helm template stable/artifactory with customCertificates enabled and confirmed the init container script includes bundle splitting logic


amarkdotdev:
Checking if this can get on a maintainer review queue soon. Much appreciated.

PR #2505, opened 2026-08-26
Provide Examples of Bridge SH Helm Chart Setup

vitaliil-jfrog:
#### PR Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated
- [ ] Variables and other changes are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. `[artifactory]`)

**What this PR does / why we need it**:
Provides Examples of Bridge SH Helm Chart Setup

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

PR #207, opened 2019-02-11
Add an option to index all builds in artifactory

danielezer:
#### PR Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [X] Chart Version bumped
- [X] CHANGELOG.md updated
- [X] Variables and other changes are documented in the README.md

**What this PR does / why we need it**:
Add an option to specify the indexAllBuilds option in xray_config.yaml (https://www.jfrog.com/confluence/display/XRAY/Configuring+Xray#ConfiguringXray-SpecifyingBuildsforAnalysis) and set it to false by default

PR #2524, opened 2026-09-01
Add Helm chart examples across 6 product charts (INST-27233)

nagarajuv-jfrog:
#### PR Checklist
- [ ] Chart Version bumped
- [ ] CHANGELOG.md updated
- [ ] Title of the PR starts with chart name (e.g. `[artifactory]`)

_This PR adds only `examples/` content — no chart code, `Chart.yaml`, or `values.yaml` in `stable/` is modified, so a version bump/changelog entry isn't applicable, matching the precedent set by #2505 and #2508._

Adds runnable Helm example configurations for `artifactory`, `artifactory-ha`, `jfrog-platform`, `catalog`, `distribution`, and `wingman`, following the `examples/<chart>/<scenario>/{README.md,<scenario>-values.yaml}` structure and style established by #2505 (Bridge) and #2508 (Worker).

Sourced from the validated Helm configuration content in jfrog/rdme-installation PR #207 and PR #209, and cross-verified against the real chart schemas in this repo — every `values.yaml` in this PR renders cleanly via `helm template` against its chart, and several were additionally spot-checked to confirm the feature actually takes effect in the rendered manifests, not just that the render succeeds.

Covers:
- **Product enablement**: Workers, Mission Control, AppTrust, Catalog, Curation, Distribution, Advanced Security (JAS), MCP Server (Wingman), AI/ML, and Artifactory Federation Service (RTFS) — across whichever of the standalone/HA/platform charts each feature applies to.
- **Licensing, TLS, custom volumes, resource/JVM sizing, and PostgreSQL password management** — both as narrow per-chart examples and, where the same feature nests differently across charts, as a consolidated `examples/advanced-helm-chart-customizations/` folder with one collapsible `<details>` section per applicable chart.
- **Ingress & networking**: Gateway API, Nginx LoadBalancer Service, SSL termination, ingress-behind-load-balancer, ingress annotations, and an external Nginx Ingress Controller in front of `jfrog-platform`.
- **Artifactory HA storage**: scaling the cluster, shared vs. per-node PersistentVolumeClaims.
- **Operational customizations**: custom init containers, sidecars, secrets, unified secret consolidation, `extraSystemYaml`, ConfigMaps, network policies, read-only root filesystem, HPA, bootstrap admin credentials, Frontend/JFBus/JFmelt deployment modes, and monitoring/logging (JMX + Filebeat).

A few real chart-schema corrections/gotchas were caught in the process (nginx TLS now required on fresh installs, several key-nesting rules between standalone and platform-chart forms, `ingress.additionalRules` needing to be a string not a list, a couple of stale/removed legacy values) — each is called out explicitly in the relevant example's README so users don't hit the same issues.

**Which issue this PR fixes**: N/A (docs/examples enhancement tracked internally as INST-27233)

This is a large PR by file count (~200 files) but each scenario is small and independent — two files each, following the existing house style exactly. Happy to split into smaller PRs per chart if preferred for easier review.


examples/artifactory-ha/custom-init-containers/custom-init-containers-values.yaml
```diff
@@ -0,0 +1,8 @@
+artifactory:
+  customInitContainersBegin: |
+    - name: "custom-setup"
+      image: alpine:3.19
```
chukka, line 4: @nagarajuv-jfrog  can we refer to echo init container image instead of random image ?

chukka (approved):

